PR #15, opened 2019-01-02
feat: Add maximum cycle checking

xxuejie:
Errors should be thrown immediately when cycle limit is reached


janx:
The `cycles` parameter in a transaction is an *exact* number instead of a maximum. It's the cycles of a transaction verification run, an error should be raised when the cycles used doesn't equal to actually used, no matter it's larger or smaller than the specified number.

xxuejie:
> The `cycles` parameter in a transaction is an _exact_ number instead of a maximum. It's the cycles of a transaction verification run, an error should be raised when the cycles specified doesn't equal to that actually used, no matter whether it's larger or smaller.

I agree `cycles` should be an exact number, however my plan is to implement the `exact` rule in CKB. Here we just use `max_cycles` to prevent infinite loop, and also be more flexible. Thoughts?

janx:
Who sets the `max_cycles` here?

xxuejie:
So for each script, `ckb` does the following parts:

1. It creates a VM instance with `max_cycles` set as the specified `cycles` for the script
2. It then runs the script in the VM
3. Upon completion, it then verifies that the actual cycles spent in the VM matches the specified `cycles` for the script

Yes I agree this is a little repetitive, but my argument here, is that the 2 checkings here serve 2 purposes:

* The `max_cycles` here set in the VM ensures no infinite loop occurs in the VM.
* CKB then ensures the cycles spent in the VM is exactly what we specified in the script, but maybe this is not a requirement needed in a different environment.

Thoughts?

janx:
I still don't see the benefits of having unnecessary maximum cycles checking here. What's the problem of checking exact cycles in ckb-vm? What's the flexibility you refer to?

xxuejie:
Actually a second thought, this is like an `YAGNI` case, let me change the code so we are checking exact cycles here.

src/lib.rs
```diff
@@ -71,11 +71,11 @@ pub enum Error {
     ParseError,
     Unaligned,
     OutOfBound,
+    IncorrectSpecifiedCycles,
```
janx: How about `InvalidCycles` or `CyclesMismatch`?

src/machine.rs
```diff
@@ -60,7 +60,7 @@ pub trait CoreMachine<R: Register, M: Memory> {
     // feature.
     fn cycles(&self) -> u64;
     fn add_cycles(&mut self, cycles: u64);
-    fn max_cycles(&self) -> Option<u64>;
+    fn available_cycles(&self) -> Option<u64>;
```
janx: `available` sounds a bit weird, `target_cycles`? thoughts?

src/machine.rs
```diff
@@ -412,12 +412,17 @@ where
...
-                if self.cycles() > max_cycles {
-                    return Err(Error::MaximumCyclesReached);
+            if let Some(available_cycles) = self.available_cycles() {
+                if self.cycles() > available_cycles {
+                    return Err(Error::IncorrectSpecifiedCycles);
                 }
             }
         }
+        if let Some(available_cycles) = self.available_cycles() {
+            if self.cycles() != available_cycles {
```
janx: Is this check redundant with line 416?

xxuejie: So line 416 checks for infinite loop, while this can help check the case where the program takes 400 cycles actually, but we specify 500 cycles. They really serve different purposes.

PR #21, opened 2019-01-16
refactor: Preparation work for moving into a JIT structure

xxuejie:
The code here currently has some obstacles that do not favor a JIT implementation:

* Overflow information is exposed outside of register value: https://github.com/nervosnetwork/ckb-vm/blob/ac2b5cf858e30a4080beffa067a8fb9e8085d097/src/instructions/register.rs#L33
* [Certain](https://github.com/nervosnetwork/ckb-vm/blob/master/src/instructions/i.rs#L227) [instruction](https://github.com/nervosnetwork/ckb-vm/blob/master/src/instructions/i.rs#L317) [logic](https://github.com/nervosnetwork/ckb-vm/blob/master/src/instructions/m.rs#L69) is implemented in Rust outside of register types, which makes it hard to JIT
* Basic CoreMachine [contains](https://github.com/nervosnetwork/ckb-vm/blob/ac2b5cf858e30a4080beffa067a8fb9e8085d097/src/machine.rs#L55-L63) too many features that are only needed in actual syscall implementation, but not in instruction implementation
* Current [registers_mut](https://github.com/nervosnetwork/ckb-vm/blob/ac2b5cf858e30a4080beffa067a8fb9e8085d097/src/machine.rs#L53) doesn't allow us to detect when a register is assigned a new value, which is needed in a JIT implementation.

This PR sorts everything out, so we can have a clean JIT structure in future PRs.

### DIFF
--- a/src/instructions/i.rs
+++ b/src/instructions/i.rs
@@ -186,13 +186,13 @@ impl Execute for Rtype {
             RtypeInstruction::SLT => {
                 let rs1_value = machine.registers()[self.rs1];
                 let rs2_value = machine.registers()[self.rs2];
-                let value = rs1_value.less_than_signed(rs2_value);
+                let value = rs1_value.lt_s(rs2_value);
                 update_register(machine, self.rd, value);
             }
             RtypeInstruction::SLTU => {
                 let rs1_value = machine.registers()[self.rs1];
                 let rs2_value = machine.registers()[self.rs2];
-                let value = rs1_value.less_than(rs2_value);
+                let value = rs1_value.lt(rs2_value);
                 update_register(machine, self.rd, value);
             }
         }
@@ -221,13 +221,13 @@ impl Execute for Itype {
             ItypeInstruction::SLTI => {
                 let rs1_value = machine.registers()[self.rs1];
                 let imm_value = R::from_i32(self.imm);
-                let value = rs1_value.less_than_signed(imm_value);
+                let value = rs1_value.lt_s(imm_value);
                 update_register(machine, self.rd, value);
             }
             ItypeInstruction::SLTIU => {
                 let rs1_value = machine.registers()[self.rs1];
                 let imm_value = R::from_i32(self.imm);
-                let value = rs1_value.less_than(imm_value);
+                let value = rs1_value.lt(imm_value);
                 update_register(machine, self.rd, value);
             }
             ItypeInstruction::JALR => {
@@ -297,10 +297,10 @@ impl Execute for Btype {
         let condition = match &self.inst {
             BtypeInstruction::BEQ => rs1_value.eq(rs2_value),
             BtypeInstruction::BNE => rs1_value.ne(rs2_value),
-            BtypeInstruction::BLT => rs1_value.less_than_signed(rs2_value),
-            BtypeInstruction::BGE => rs1_value.greater_or_equal_than_signed(rs2_value),
-            BtypeInstruction::BLTU => rs1_value.less_than(rs2_value),
-            BtypeInstruction::BGEU => rs1_value.greater_or_equal_than(rs2_value),
+            BtypeInstruction::BLT => rs1_value.lt_s(rs2_value),
+            BtypeInstruction::BGE => rs1_value.ge_s(rs2_value),
+            BtypeInstruction::BLTU => rs1_value.lt(rs2_value),
+            BtypeInstruction::BGEU => rs1_value.ge(rs2_value),
         };
         let next_pc_offset = condition.cond(R::from_i32(self.imm), R::from_usize(4));
         Ok(Some(machine.pc().overflowing_add(next_pc_offset)))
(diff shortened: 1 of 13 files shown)


src/instructions/register.rs
```diff
@@ -26,31 +21,49 @@ pub trait Register:
...
     fn max_value() -> Self;
 
-    // By default, PartialOrd/Ord compares unsigned value, which is the same
-    // as shl/shr.
-    fn signed_cmp(&self, other: &Self) -> Ordering;
+    // Conditional operations, if the condition evaluated here is true, R::one()
+    // will be emitted, otherwise R::zero() will be emitted
+    fn eq(self, other: Self) -> Self;
+    fn less_than(self, other: Self) -> Self;
+    fn less_than_signed(self, other: Self) -> Self;
```
TheWaWaR: suggestion: `eq()`, `lt()`, `lt_signed()`.

src/instructions/register.rs
```diff
@@ -26,31 +21,49 @@ pub trait Register:
...
+    // | Condition              | Dividend  | Divisor | DIVU[W] | REMU[W] |  DIV[W]   | REM[W] |
+    // +------------------------+-----------+---------+---------+---------+-----------+--------+
+    // | Division by zero       |     x     |    0    | 2**L-1  |    x    |    -1     |   x    |
+    // +------------------------+-----------+---------+---------+---------+-----------+--------+
+    // | Overflow (signed only) | −2**(L−1) |   −1    |    -    |    -    | -2**(L-1) |   0    |
+    // +---------------------------------------------------------------------------------------+
+    fn overflowing_div(self, rhs: Self) -> Self;
+    fn overflowing_rem(self, rhs: Self) -> Self;
+    fn overflowing_div_signed(self, rhs: Self) -> Self;
+    fn overflowing_rem_signed(self, rhs: Self) -> Self;
```
TheWaWaR: Why no overflow indicator returned?

src/instructions/i.rs
```diff
@@ -298,27 +292,18 @@ impl Execute for Btype {
...
-            BtypeInstruction::BGEU => {
-                machine.registers()[self.rs1] >= machine.registers()[self.rs2]
-            }
+        let rs1_value = machine.registers()[self.rs1];
+        let rs2_value = machine.registers()[self.rs2];
+        let condition = match &self.inst {
+            BtypeInstruction::BEQ => rs1_value.eq(rs2_value),
+            BtypeInstruction::BNE => rs1_value.eq(rs2_value).logical_not(),
+            BtypeInstruction::BLT => rs1_value.less_than_signed(rs2_value),
+            BtypeInstruction::BGE => rs1_value.less_than_signed(rs2_value).logical_not(),
```
zhangsoledad:
```
pub trait PartialEq<Rhs: ?Sized = Self> {
  #[must_use]
  fn eq(&self, other: &Rhs) -> bool;
  #[must_use]
  fn ne(&self, other: &Rhs) -> bool { ... }
}
```
Maybe we can follow `PartialEq`, provide `ne` method?

src/instructions/register.rs
xxuejie: We want all the returned information here to be contained within Register type, so we can provide enough information for JIT(if we have returned `overflow` indicator here, and use it in a Rust conditional in `instructions` package, JIT might not have enough information to transform code). And since those div/rem instructions are the only place where we actually use overflow indicators, we can change the implementation to put div/rem logic within registers, and expose no overflow information.

src/instructions/register.rs
```diff
@@ -94,6 +94,18 @@ pub trait Register:
...
+        self.eq(rhs).logical_not()
+    }
+
+    fn greater_or_equal_than(self, other: Self) -> Self {
+        self.less_than(other).logical_not()
+    }
+
+    fn greater_or_equal_than_signed(self, other: Self) -> Self {
+        self.less_than_signed(other).logical_not()
+    }
```
TheWaWaR: How about:
`eq`, `ne`, `gt`, `ge`, `lt`, `le`
https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html

xxuejie: Is it weird that `le` is in abbreviation, but `signed` is in full form?

TheWaWaR: Yes, but `eq` vs `less_than` is also wired

xxuejie: Which actually brings my second question: should we go with `eq`, `lt`, `lt_signed`, or `equal`, `less_than`, `less_than_signed`?

zhangsoledad: `le_s`, `ge_s`, not bad

doitian: `greater_or_equal_than` is weird. Usually, we omit the first preposition and get `greater_or_equal_to`, or use the full form `greater_than_or_equal_to`.

Another reason to use abbreviations.

doitian: Many languages use the abbreviations, such as Python, Lua

PR #24, opened 2019-01-24
refactor: change Syscalls and DefaultMachine generic type to SupportMachine

quake:


src/machine.rs
```diff
@@ -222,124 +222,89 @@ impl<R: Register, M: Memory> SupportMachine for DefaultCoreMachine<R, M> {
...
 
-    fn set_register(&mut self, idx: usize, value: R) {
-        self.registers[idx] = value;
+    fn set_register(&mut self, idx: usize, value: Self::REG) {
+        self.core.set_register(idx, value)
     }
 }
 
-impl<'a, R: Register, M: Memory> SupportMachine for DefaultMachine<'a, R, M> {
+impl<Core: SupportMachine> SupportMachine for DefaultMachine<'_, Core> {
```
xxuejie: Can we rename `Core` here to something else, maybe `Mac`? It feels a little weird that `Core` implements `SupportMachine`, while we also have `CoreMachine`

src/machine.rs
```diff
@@ -222,124 +222,89 @@ impl<R: Register, M: Memory> SupportMachine for DefaultCoreMachine<R, M> {
...
     }
 
     fn cycles(&self) -> u64 {
-        self.cycles
+        self.core.cycles()
     }
 
     fn set_cycles(&mut self, cycles: u64) {
-        self.cycles = cycles;
+        self.core.set_cycles(cycles)
```
xxuejie: `;` is missing here

src/machine.rs
```diff
@@ -222,124 +222,89 @@ impl<R: Register, M: Memory> SupportMachine for DefaultCoreMachine<R, M> {
...
-impl<'a, R: Register, M: Memory> SupportMachine for DefaultMachine<'a, R, M> {
+impl<Core: SupportMachine> SupportMachine for DefaultMachine<'_, Core> {
     fn elf_end(&self) -> usize {
-        self.elf_end
+        self.core.elf_end()
     }
 
     fn set_elf_end(&mut self, elf_end: usize) {
-        self.elf_end = elf_end;
+        self.core.set_elf_end(elf_end)
```
xxuejie: `;` is missing

doitian:
I'm a bit curious why we call it `SupportMachine`?

xxuejie:
> I'm a bit curious why we call it `SupportMachine`? It lets me recall Support Vector Machine.

It's called `SupportMachine` since it provides supporting functions needed in syscalls which are not exactly components a physical machine has. In that sense, yes, we can also call it `SupportVectorMachine`, but the other question is `vector` has its own special meaning in low level machine, that might bring some confusion.

src/machine.rs
quake: It intentionally omits `;` here, it's a delegation call to inner.

PR #36, opened 2019-03-24
chore: Make JIT into a feature that can be enabled or disabled

xxuejie:
This way we can upgrade CKB with latest CKB-VM containing bug fixes, but don't need to worry that JIT related code could bring potential problems for now.


xxuejie:
The additional CI items shown here seem to be a problem with Travis itself, the actual CI tests have passed.

PR #37, opened 2019-03-25
Use traces to optimize interpreter performance

xxuejie:
Note this is actually at the same level of performance comparing to our currently deployed [version](https://github.com/nervosnetwork/ckb-vm/tree/simple-decoder-cache-64-2) using decoder cache. But the advantage of this solution here, is that we are using fixed length vector instead of a dynamically grown HashMap, hence we will have a fixed upper bound on the memory usage when running any arbitrary scripts.

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -428,6 +428,9 @@ impl<'a, Inner: SupportMachine> DefaultMachine<'a, Inner> {
                 traces[slot].address = pc;
             }
             for i in &traces[slot].instructions {
+                if let Instruction::Empty = i {
+                    break;
+                }
                 i.execute(self)?;
                 let cycles = self
                     .instruction_cycle_func
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -485,3 +485,17 @@ impl<'a, Inner> DefaultMachineBuilder<'a, Inner> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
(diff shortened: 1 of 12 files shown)


src/machine.rs
```diff
@@ -342,6 +344,22 @@ impl<Inner: CoreMachine> Display for DefaultMachine<'_, Inner> {
...
 }
 
+// The number of trace items to keep
+const TRACE_SIZE: usize = 8192;
+// Quick bit-mask to truncate a value in trace size range
+const TRACE_MASK: usize = (TRACE_SIZE - 1);
+// The maximum number of instructions to cache in a trace item
+const TRACE_ITEM_LENGTH: usize = 16;
+// Shifts to truncate a value so 2 traces won't share the same code,
+// should be log2(TRACE_ITEM_LENGTH)
```
doitian: Better add a test for this assumption.

src/machine.rs
```diff
@@ -383,19 +401,44 @@ impl<'a, Inner: SupportMachine> DefaultMachine<'a, Inner> {
...
+                    traces[slot].instructions[i] = instruction;
+                    if end_instruction {
+                        break;
+                    }
+                    i += 1;
+                }
+                traces[slot].address = pc;
+            }
+            for i in &traces[slot].instructions {
+                if let Instruction::Empty = i {
```
doitian: If Empty is only used as a placeholder here, I would suggest just use Option<Instruction> in the trace.

PR #31, opened 2019-02-26
refactor: use bits operations to round up/down

doitian:
```
rounup via remainder    time:   [2.2321 us 2.2439 us 2.2577 us]
                        change: [-5.5708% -3.8512% -2.2337%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe

rounup via bit ops      time:   [0.0000 ps 0.0000 ps 0.0000 ps]
                        change: [-53.351% -6.3284% +86.775%] (p = 0.87 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  4 (4.00%) high mild
  8 (8.00%) high severe

rounup via multication  time:   [2.4807 us 2.5208 us 2.5662 us]
                        change: [+4.8590% +7.3068% +9.9176%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 11 outliers among 100 measurements (11.00%)
  8 (8.00%) high mild
  3 (3.00%) high severe

rounup via remainder    time:   [2.3393 us 2.3603 us 2.3835 us]
                        change: [+5.4361% +7.3216% +9.4704%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe

rounup via bit ops      time:   [0.0000 ps 0.0000 ps 0.0000 ps]
                        change: [-46.005% +9.5582% +124.75%] (p = 0.82 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  4 (4.00%) high mild
  8 (8.00%) high severe

rounup via multication  time:   [2.3288 us 2.3491 us 2.3713 us]
                        change: [-10.144% -8.3477% -6.5150%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
```

PR #38, opened 2019-03-26
Rc/v0.8.0

xxuejie:
Features:

* Baseline JIT (#29)

Refactors:

* use bits operations to round up/down (#31)
* Use Pin in Rust 1.33 to simplify code  (#32) 
* use `[ e; n ]` to init array (#33)
* prefer NonNull to store raw pointer (#34)
* Make JIT into a feature that can be enabled or disabled (#36)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ckb-vm"
-version = "0.1.0"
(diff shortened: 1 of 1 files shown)


doitian:
v0.8.0 should already have been released.

PR #41, opened 2019-04-01
Instruction structure refactoring

xxuejie:
This is a major refactoring work for the internal Instruction data structure, the revised structure has 2 benefits:

1. It takes only 8 bytes instead of the original 16 bytes, meaning we will be using less memory here, and more trace items will be available for us to further boost performance.
2. The revised instruction structure is also optimized for fast decoding, which enables the possibility of writing the hot interpreter loop in pure assembly to make it even faster.


src/instructions/mod.rs
```diff
@@ -7,184 +8,364 @@ pub mod m;
...
+// instruction into the internal form here(much like how traces/micro-ops work.)
+pub type Instruction = u64;
+
+pub type InstructionOpcode = u8;
+
+// ADDI must have an opcode of 0, this way the default 0 value for u64
+// happens to be a NOP operation in RISC-V format. Of course using ADD here
+// should have the same effect as well but we just want to stick to RISC-V
+// established convention here.
+pub const OP_ADDI: InstructionOpcode = 0;
```
doitian: Ever tried declaring an enum with `repr(u8)`?

xxuejie: Just did some experiment:

* [The original unoptimized case](https://github.com/xxuejie/ckb-vm/blob/d86d1ae8e9d617ecef467d52a099713650655db8/src/instructions/mod.rs#L44): 32.242 ms
* [Adding repr(u8) alone](https://github.com/xxuejie/ckb-vm/blob/3c0675a4194d9e6150df9cbe443924c8a66114ab/src/instructions/mod.rs#L45): 32.519 ms
* [Using repr(u8) and transmute](https://github.com/xxuejie/ckb-vm/blob/c2a9303510f594490d93050806e4c37f75297a43/src/instructions/mod.rs#L46): 30.875 ms
* [Using just plain u8 constants](https://github.com/xxuejie/ckb-vm/blob/5838875739396217be73a545545105bc480ac654/src/instructions/mod.rs#L47): 27.736 ms

My guess here is that `[repr(u8)]` really just only affects the enum size here, and since we already make sure the enum is an [u8](https://github.com/xxuejie/ckb-vm/blob/d86d1ae8e9d617ecef467d52a099713650655db8/src/instructions/mod.rs#L442), it doesn't really make much difference for performance.

PR #45, opened 2019-04-17
feat: add assembly based interpreter loop to speeeup VM executions

xxuejie:
This PR adds a new assembly based interpreter loop to speed the VM in x86_64 architectures. While running our [secp256k1-sha3 benchmark](https://github.com/nervosnetwork/ckb-vm-bench-scripts/blob/master/c/secp256k1_bench.c)

* Pure Rust based interpreter: ~60ms
* Rust based interpreter with trace optimization: ~27ms
* Baseline JIT: ~9ms
* This PR: ~7.9ms

Note there's still room for more optimizations(profiling shows current work has a relatively higher portion of branch mis-predictions, which can affect the performance a lot), but this PR is quite large and let's leave more optimization work to the future.


quake:
It's awesome but I don't understand assembly code, let's test it with ckb :)

doitian:
TODO: release it to CKB and deploy it to test

PR #46, opened 2019-04-18
fix: flat memory implementation stores byte in wrong locations

lerencao:
As far as I can know, `slice_ptr` should start from `addr`.


xxuejie:
Thanks!

PR #55, opened 2019-05-15
feat: Lazy memory initialization in assembly based implementation

xxuejie:
This saves an unnecessary memory zeroing process, in the offline test,
this only brings ~ 0.3ms speedup, but in a multi-threaded environment
where multiple VM instances are running in parallel, this might make a difference


xxuejie:
Closing this since #56 brings bigger performance boost, and deploying this after #56 will actually decrease performance.

PR #56, opened 2019-05-17
perf: Numerous performance tweaks based on current design assumption

xxuejie:
On a multi-threaded test environment, this can bring 20% - 25%
performance boost

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -8,7 +8,7 @@ use crate::{
         check_permission, fill_page_data, memset, FLAG_EXECUTABLE, FLAG_FREEZED, FLAG_WRITABLE,
     },
     CoreMachine, DefaultMachine, Error, Machine, Memory, SupportMachine, RISCV_MAX_MEMORY,
-    RISCV_PAGESIZE,
+    RISCV_PAGES, RISCV_PAGESIZE,
 };
 use byteorder::{ByteOrder, LittleEndian};
 use bytes::Bytes;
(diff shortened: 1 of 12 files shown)


src/memory/sparse.rs
```diff
@@ -79,6 +78,14 @@ impl<R: Register> Memory<R> for SparseMemory<R> {
         fill_page_data(self, addr, size, source, offset_from_addr)
     }
 
+    fn fetch_flag(&mut self, page: usize) -> Result<u8, Error> {
+        if page < RISCV_MAX_MEMORY / RISCV_PAGESIZE {
```
quake: replace all `RISCV_MAX_MEMORY / RISCV_PAGESIZE` to `MAX_PAGES`
```suggestion
        if page < MAX_PAGES {
```

src/machine/asm/mod.rs
```diff
@@ -76,36 +76,39 @@ impl Memory<u64> for Box<AsmCoreMachine> {
             if self.flags[page] & FLAG_FREEZED != 0 {
                 return Err(Error::InvalidPermission);
             }
+        }
+        fill_page_data(self, addr, size, source, offset_from_addr)?;
+        for page_addr in (addr..addr + size).step_by(RISCV_PAGESIZE) {
```
quake:
```suggestion
for page in (addr..(addr + size - 1) / page_size + 1) {
  self.flags[page] = flags;
}
```

xxuejie: Comment added explaining the code, the suggested change is actually slower in a Rust setup.

PR #59, opened 2019-05-27
chore: fix a bad way to using machine

mohanson:
The old codes:

```rs
let value = &machine.registers()[i.rs2()]; // immutable borrow machine
update_register(machine, i.rd(), value.clone()); // mutable borrow machine in first argument, and immutable borrow in third argument "value"
```

### DIFF
--- a/src/instructions/execute.rs
+++ b/src/instructions/execute.rs
@@ -639,8 +639,8 @@ pub fn execute<Mac: Machine>(inst: Instruction, machine: &mut Mac) -> Result<(),
         }
         insts::OP_RVC_MV => {
             let i = Rtype(inst);
-            let value = &machine.registers()[i.rs2()];
-            update_register(machine, i.rd(), value.clone());
+            let value = machine.registers()[i.rs2()].clone();
+            update_register(machine, i.rd(), value);
             None
         }
         insts::OP_RVC_JAL => {


src/instructions/execute.rs
```diff
@@ -639,8 +639,8 @@ pub fn execute<Mac: Machine>(inst: Instruction, machine: &mut Mac) -> Result<(),
         }
         insts::OP_RVC_MV => {
             let i = Rtype(inst);
-            let value = &machine.registers()[i.rs2()];
-            update_register(machine, i.rd(), value.clone());
+            let value = machine.registers()[i.rs2()].clone();
+            update_register(machine, i.rd(), value);
```
xxuejie: I'm not sure if the code here makes a difference, `value.clone()` part in the old code already kills the immutable borrow, so the second line here only has a mutable borrow on the machine type.

Personally I have no problems writing the code either way, but the point I want to make is: if we do want to make the change, there're many other places in this module where the code is written in a similar style, so we either keep them unchanged, or change all of the code together to keep a unified style.

Thoughts?

mohanson: I think this is the only place in this module that needs to be modified. 

Rust 1.35.0 stable gives warnings see bellows:

```
warning: cannot borrow `*machine` as mutable because it is also borrowed as immutable
   --> src\instructions\execute.rs:643:29
    |
642 |             let value = &machine.registers()[i.rs2()];
    |                          ------- immutable borrow occurs here
643 |             update_register(machine, i.rd(), value.clone());
    |                             ^^^^^^^          ----- immutable borrow later used here
    |                             |
    |                             mutable borrow occurs here
    |
    = note: #[warn(mutable_borrow_reservation_conflict)] on by default
    = warning: this borrowing pattern was not meant to be accepted, and may become a hard error in the future
    = note: for more information, see issue #59159 <https://github.com/rust-lang/rust/issues/59159>
```

xxuejie: Ah I see, I didn't know this behavior change, in this case this is all good, I still want to change all the other parts as a unified style change but we can leave it till another time. Thanks

PR #61, opened 2019-05-27
chore: add an example named is13

mohanson:
Add an example named "is13" to shows how to build C source code to riscv output, and then run this ouput by ckb-vm.

```
$ cargo run --example is13 13    # b"13" is thirteen
$ cargo run --example is13 0xd   # b"0xd" is thirteen
$ cargo run --example is13 HELLO # b"HELLO" is not thirteen
```

# Why I add this?

I'm a newer for riscv or ckb-vm, When I first saw this project, I almost didn’t know how to get start. So, I googled, searched in QUIP, or some nervos public articles, which takes me so much time. It's worth to add an example in this repo.


examples/is13.rs
```diff
@@ -0,0 +1,45 @@
...
+//
+//   $ git clone --recursive https://github.com/riscv/riscv-gnu-toolchain
+//   $ cd riscv-gnu-toolchain
+//   $ mkdir build && cd build
+//   $ ../configure --prefix=/opt/riscv --with-arch=rv32imac --with-abi=ilp32
+//   $ make
+//
+// On my ubuntu machine, it takes 1 hours. Sad.
+//
+// Then, you can some c codes like is13.c, and build it by "riscv32-unknown-elf-gcc"
```
doitian: Not a sentence

examples/is13.rs
```diff
@@ -0,0 +1,45 @@
...
+//
+//   $ riscv32-unknown-elf-gcc -o is13 is13.c
```
doitian: I think we don't need to commit the binary file into the repository since it's an example to let the user go through the whole process.

examples/is13.rs
mohanson: Fixed.

examples/is13.rs
mohanson: Removed it. It should be done by the user himself, emmm.

doitian: Please add it to `.gitignore` as well

mohanson: Added!

PR #62, opened 2019-05-30
fix: VM panics when ELF uses invalid file offset

xxuejie:
This is a bug reported by @rink1969. When an ELF file uses invalid file offset, current VM implementation would panic.


janx:
Will ckb process crash on vm crash?

The ckb process should be designed to not crash on vm crash, and just return false as transaction verification result. With this protection a vm panic can only impact a single tx.

xxuejie:
> Will ckb process crash on vm crash?
> 
> The ckb process should be designed to not crash on vm crash, and just return false as transaction verification result. With this protection a vm panic can only impact a single tx.

Yes CKB process will crash on VM crash since right now VM is just a library running together with CKB, however this is really a bug, not a desired behavior. It remains a question if we need to make VM a separate process.

PR #63, opened 2019-05-30
fix: out of bound read check in assembly VM

xxuejie:


xxuejie:
@nervos-bot give me five

PR #65, opened 2019-06-05
test: add code coverage

u2:
https://github.com/nervosnetwork/ckb/pull/943

Using kcov instead of tarpaulin, because it's for compiled programs not only Rust, and we can also compare with tarpaulin for a while.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,36 @@ matrix:
             - git
             - build-essential
       env: SUITE=ci-all-features
+    - name: Code Coverage
+      rust: 1.33.0
+      sudo: true
+      before_install:
+        - sudo apt-get update
+      addons:
+        apt:
+          packages:
+            - git
+            - build-essential
+            - libcurl4-openssl-dev
+            - libelf-dev
+            - libdw-dev
+            - cmake
+            - gcc
+            - binutils-dev
+      script:
+        wget https://github.com/SimonKagstrom/kcov/archive/v36.tar.gz &&
+        tar xzf v36.tar.gz &&
+        cd kcov-36 &&
+        mkdir build &&
+        cd build &&
+        cmake .. &&
+        make &&
+        sudo make install &&
+        cd ../.. &&
+        rm -rf kcov-36 v36.tar.gz &&
+        make cov &&
+        bash <(curl -s https://codecov.io/bash) &&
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -50,6 +50,35 @@ matrix:
...
+            - git
+            - build-essential
+            - libcurl4-openssl-dev
+            - libelf-dev
+            - libdw-dev
+            - cmake
+            - gcc
+            - binutils-dev
+      script:
+        wget https://github.com/SimonKagstrom/kcov/archive/master.tar.gz &&
```
xxuejie: Can we lock a specific release? In my experience tools working on binary levels are usually ... flaky, and it might resolve in unwanted error in a total different PR if upstream is updated somehow.

doitian: We can download zip from [releases](https://github.com/SimonKagstrom/kcov/releases)

.travis.yml
```diff
@@ -50,6 +50,35 @@ matrix:
             - git
             - build-essential
       env: SUITE=ci-all-features
+    - rust: 1.33.0
```
doitian: Please give this job a name

```suggestion
    - name: Code Coverage
      rust: 1.33.0
```

.travis.yml
u2: updated

u2:
https://travis-ci.com/nervosnetwork/ckb-vm/jobs/205521537#L1505

The ubuntu 14 file matching is different from ubuntu 16? It works well on my computer.

xxuejie:
I'm taking another look at this, and I feel like adding the coverage code here might not be a good solution. For space reason, this repo here only contains a few basic tests for sanity checks as well as any tests discovered by bugs. A more complete test on the VM resides at https://github.com/nervosnetwork/ckb-vm-test-suite. So I feel it's better we also put coverage code there.

u2:
> I'm taking another look at this, and I feel like adding the coverage code here might not be a good solution. For space reason, this repo here only contains a few basic tests for sanity checks as well as any tests discovered by bugs. A more complete test on the VM resides at https://github.com/nervosnetwork/ckb-vm-test-suite. So I feel it's better we also put coverage code there.

So, I move the code coverage to the test suite repo?

PR #68, opened 2019-06-05
Replace uses of rounddown / roundup with round_page_down / round_page_up

brson:
There was already an implementation of round_page_down that duplicated rounddown. This cleanup patch adds round_page_up and makes all rounddown / roundup uses call the round_page versions.


brson:
The CI failures due to me not testing with various features enabled. I'll fix it in the next few days.

brson:
@xxuejie CI failures fixed.

PR #71, opened 2019-06-10
Eliminate as much usize as making sense from CKB VM

xxuejie:
`usize` is a tricky part since it has different widths across 32-bit vs 64-bit architectures. This PR tries to remove the use of usize from the parts that are making sense. After this PR, we should only be using usize where we have to(such as indexing arrays), and constants where we know no overflows would occur.

### DIFF
--- a/src/jit/machine.rs
+++ b/src/jit/machine.rs
@@ -54,14 +54,14 @@ struct RustData {
     freezed: bool,
     // First item in the value tuple is starting address of jitted basic block,
     // second item in the tuple is current cycle count of basic block.
-    jitted_blocks: FnvHashMap<usize, (usize, u64)>,
+    jitted_blocks: FnvHashMap<u64, (u64, u64)>,
     // This records the start address(first item) and length(second item) of all
     // jitted basic block, so when a memory write happens within a jitted block,
     // we can use this as a hint to clear those jitted blocks since they will be
     // incorrect afther the memory write. Notice this vec is also sorted based on
     // the start address(first item) to allow for binary search operations for speed
     // considerations.
-    jitted_segments: Vec<(usize, usize)>,
+    jitted_segments: Vec<(u64, u64)>,
     // We cannot use generics here since the FFI functions at the end of this
     // file need to stay unmangled, but Rust requires mangled names for
     // functions with generic types.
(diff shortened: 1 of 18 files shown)


xxuejie:
The failed CI check is due to codecov mis-configuration. I've revised the configuration, and it should work all right in future PRs.

src/machine/asm/mod.rs
```diff
@@ -76,110 +78,121 @@ impl Memory<u64> for Box<AsmCoreMachine> {
...
-    fn execute_load16(&mut self, addr: usize) -> Result<u16, Error> {
+    fn execute_load16(&mut self, addr: u64) -> Result<u16, Error> {
         check_permission(self, addr, 2, FLAG_EXECUTABLE)?;
         self.load16(&(addr as u64)).map(|v| v as u16)
     }
 
     fn load8(&mut self, addr: &u64) -> Result<u64, Error> {
-        let addr = *addr as usize;
-        if addr + 1 > self.memory.len() {
+        let addr = *addr as u64;
```
doitian: Many redudant `as u64` in this file.

xxuejie: Ooops I was relying too much on clippy, already resolved.

src/jit/machine.rs
```diff
@@ -284,7 +284,7 @@ impl BaselineJitMachine {
                 } else {
                     // TODO: check if dynasm handles alignments here.
                     emitter.encode(&mut buffer_mut[rust_data.used..])?;
-                    let offset = rust_data.used;
+                    let offset = rust_data.used as u64;
```
jjyr: I notice the `used` field is still usize type, not sure if this is missed or is intentioned.

xxuejie: Yes this is kept as usize for 2 reasons:

1. Actually [here](https://github.com/nervosnetwork/ckb-vm/blob/2ffd6f12b750949c88103f26eb2b27e762ff1d8f/src/jit/machine.rs#L281) we ensure that used can never exceed 1MB, which will fit well within usize even for 32-bit platform
2. The whole JIT package right now is still treated as a PoC and won't be included in production, in fact, I'm thinking of removing the whole JIT package once the new AOT mode is in place, that's why this PR only contains necessary fixes in the JIT package, and still many fields in JIT package as usize.

PR #72, opened 2019-06-13
feat: Add new AOT mode to replace experimental JIT mode

xxuejie:
This PR adds a new AOT mode evolved from previously experimental JIT mode. It works by compiling a RISC-V program directly into x64 assemblies. After that, most of the code can just run in the compiled x64 assembly binary without any overhead.

Benchmark shows that this can further bump VM performance from 6ms measured in assembly interpreter to ~1.5ms(of which ~0.9ms is the actual running time).

With this new AOT mode, we are also removing our old JIT mode which serves its purpose.


src/machine/aot/mod.rs
```diff
@@ -0,0 +1,682 @@
...
+            // A section must end a basic block, otherwise we would run into
+            // out of bound error;
+            if !start_of_basic_block {
+                return Err(Error::OutOfBound);
+            }
+            debug_assert!(!self.labels.contains(&section_end));
+        }
+        // Remove all labels pointed to dummy sections, since we won't generate
+        // code for dummy sections
+        self.labels = self.dummy_sections.iter().fold(
```
doitian: See `HashSet::retain`

src/machine/aot/mod.rs
```diff
@@ -0,0 +1,682 @@
...
+            self.pc = Value::from_u64(section_start);
+            let mut start_of_basic_block = true;
+            loop {
+                let pc = self.read_pc()?;
+                if pc >= section_end {
+                    break;
+                }
+                match decoder.decode(&mut self.memory, pc) {
+                    Ok(instruction) => {
+                        start_of_basic_block = is_basic_block_end_instruction(instruction);
```
doitian: The whole loop looks weird. I think the reason is that in the current iteration, we really are processing the instruction in the next iteration.

doitian: Pseudo code

```python
pc = section_start
start_of_basic_block = true

while pc < section_end:
    if start_of_basic_block:
        labels.append(pc)

    instruction = decode_at(pc) 
    execute(instruction)
    scan_labels_to_test()
    start_of_basic_block = is_end_of_basic_block(instruction)
    pc = pc + instruction_length(instruction)
```

src/machine/aot/mod.rs
xxuejie: This is actually a deliberate choice. If we use `HashSet::retain`, the outer loop would be looping through labels, then for each label we check against all the sections. That means for `n` labels and `m` sections, we need `n*m` checks.

The code here actually has its outer loop looping through sections, then in each section, we loop through remaining labels, that means in each iteration on different sections, we might be dealing with less labels hence less checks.

Note this is not yet benchmarked, it's just due to mental thoughts. I can still change it if you have other thoughts.

src/machine/aot/aot.x64.c
```diff
@@ -0,0 +1,1284 @@
...
+  | lea rdx, machine->memory
+  switch (size) {
+    case 1:
+      | movzx rcx, byte [rdx+rax]
+      break;
+    case 2:
+      | movzx rcx, word [rdx+rax]
+      break;
+    case 4:
+      | mov ecx, dword [rdx+rax]
```
rink1969: why not use rcx like other case？

xxuejie: In x64 assembly when writing to `ecx`, the higher 32-bit of `rcx` will automatically be cleared. So using `mov ecx, dword [rdx+rax]` here has the same effect as `mov rcx, dword [rdx+rax]`, but with the benefit of one less bytes in the encoded assembly, which is nicer to the CPU frontend decoder.

That being said, we should change the code to use `ecx` in all cases here except when we are writing all 64 bits.

src/machine/aot/aot.x64.c
```diff
@@ -0,0 +1,1284 @@
...
+      break;
+    case AOT_TAG_IMMEDIATE:
+      | load_imm rcx, b.value.i
+      | mul rcx
+      break;
+    case AOT_TAG_X64_REGISTER:
+      | mul Rq(b.value.x64_reg)
+      break;
+  }
+  /* calculate ~res and store it in rcx */
```
rink1969: There is no rcx in below code.
Is this comment correct?

xxuejie: Oops this is something I forgot to delete

doitian:
- [ ] Review `LabelGatheringMachine`
- [ ] Review `AotCompilingMachine`
- [ ] Review `aot.x64.c`

src/machine/aot/aot.x64.c
```diff
@@ -0,0 +1,1283 @@
...
+          | mov qword machine->pc, rcx
+          break;
+        case 0x40:
+          | mov qword machine->pc, ((uint32_t)(value.value.i & 0x7FFFFFFF))
+          | jmp =>((value.value.i >> 32) ^ 0x40000000)
+          break;
+        case 0x0:
+          | load_imm rcx, value.value.i
+          | mov machine->pc, rcx
+          ret = aot_exit(context, CKB_VM_ASM_RET_DYNAMIC_JUMP);
```
doitian: Is this an unexpected jump in a code block, so we return an error here?

src/machine/aot/mod.rs
```diff
@@ -0,0 +1,673 @@
...
+        for i in instructions {
+            let pc = self.read_pc()?;
+            let length = instruction_length(*i);
+            cycles += self
+                .instruction_cycle_func
+                .as_ref()
+                .map(|f| f(*i))
+                .unwrap_or(0);
+            execute(*i, self)?;
+            self.pc = Value::from_u64(pc + u64::from(length));
```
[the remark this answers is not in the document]
xxuejie: Notice here it's encoding time, not running time. We are merely translating each instruction in sequence. The actual pc could jump to different places

src/machine/aot/aot.x64.c
xxuejie: I suggest we consider the return here as an indicator that the current x64 code doesn't know where to go next, hence it would exit first , and use Rust code's aid to figure out the next instruction to execute, the code here does exactly that. This could due to a jump to a place where no label is generated, or a jump based on a runtime value in a register

src/machine/aot/aot.x64.c
```diff
@@ -0,0 +1,1283 @@
...
+      switch ((uint8_t) (value.value.i >> 56)) {
+        case 0x80:
+          /*
+           * This means just writing the result without actually jump
+           */
+          | load_imm rcx, (value.value.i & 0xFFFFFFFFFFFFFF)
+          | mov qword machine->pc, rcx
+          break;
+        case 0x40:
+          | mov qword machine->pc, ((uint32_t)(value.value.i & 0x7FFFFFFF))
```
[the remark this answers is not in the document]
xxuejie: Dynasm leverages signed 32 bit integer a lot internally(including the immediate value we are setting here), I'm merely being more carful and obey the rules. Besides 7FFFFFFF is already much more than the maximum memory used in CKB VM

src/machine/aot/aot.x64.c
doitian: I see. I thought it returns the code as the final VM script result, indeed the exit code is passed to the `match result` block in `AsmMachine::run`.

src/machine/aot/mod.rs
```diff
@@ -0,0 +1,673 @@
...
+
+    pub fn gather(&mut self) -> Result<(), Error> {
+        let decoder = build_imac_decoder::<u64>();
+        for i in 0..self.sections.len() {
+            let (section_start, section_end) = self.sections[i];
+            self.pc = Value::from_u64(section_start);
+            let mut start_of_basic_block = true;
+            while self.read_pc()? < section_end {
+                let pc = self.read_pc()?;
+                if start_of_basic_block {
```
doitian: I think should be moved `Ok(instruction)` branch, otherwise, it will add dummy section to labels.

xxuejie: I think the result is the same since labels for dummy section will be cleaned, but I agree moving this will make the logic cleaner.

src/machine/aot/mod.rs
doitian:
```
        for (dummy_start, dummy_end) in &self.dummy_sections {
            self.labels.retain(|label| {
                *label < *dummy_start || *label >= *dummy_end
            })
        }
```

src/machine/aot/mod.rs
```diff
@@ -0,0 +1,673 @@
...
+            .rev()
+            .collect();
+        // Test there's no empty section
+        if sections.iter().any(|(s, e)| s >= e) {
+            return Err(Error::OutOfBound);
+        }
+        // Test no section overlaps with one another. We first sort section
+        // list by start, then we test if each end is equal or less than
+        // the next start.
+        sections.sort_by_key(|section| section.0);
```
doitian: Just some quick question, not code change suggestions.

1. Can code jump between sections?
2. If so, will sorting sections help improve the performance a bit? Since the RISC-V code may already have optimized the code to increase the chance to do sequential read on jumping. If we keep the same sequence in the x86 ASM, we may benefit from this optimization.

src/machine/aot/mod.rs
```diff
@@ -0,0 +1,673 @@
...
+            return Err(Error::OutOfBound);
+        }
+        self.emitter.emit(&Write::Pc {
+            value: Value::Imm(pc | ADDRESS_WRITE_ONLY_FLAG),
+        })?;
+        for write in &self.writes {
+            self.emitter.emit(&write)?;
+        }
+        self.writes.clear();
+        if let Some(value) = self.next_pc_write.take() {
```
doitian: Should we check value is not out of bound here?

src/machine/aot/mod.rs
```diff
@@ -0,0 +1,673 @@
...
+            execute(*i, self)?;
+            self.pc = Value::from_u64(pc + u64::from(length));
+        }
+        let pc = self.read_pc()?;
+        self.emitter.emit_add_cycles(cycles)?;
+        // Emit succeeding PC write only
+        if pc >= RISCV_MAX_MEMORY as u64 {
+            return Err(Error::OutOfBound);
+        }
+        self.emitter.emit(&Write::Pc {
```
doitian: Is it safe to move this to the else branch of `if let Some(value) = self.next_pc_write.take() {`.

src/machine/aot/mod.rs
xxuejie: 1. Yes code can jump freely between sections
2. It's hard to say if this is gonna bring any improvement, normal GCC outputs actually only have one text section for all the code, and code reorganization happens within this single section.

src/machine/aot/mod.rs
```diff
@@ -0,0 +1,673 @@
...
+    }
+}
+
+pub struct AotCompilingMachine {
+    registers: [Value; 32],
+    pc: Value,
+    emitter: Emitter,
+    memory: FlatMemory<u64>,
+    sections: Vec<(u64, u64)>,
+    dummy_sections: HashMap<u64, u64>,
```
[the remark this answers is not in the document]
xxuejie: The problem is we don't know if this is all due to padding, could also be just some data filled in text section

src/machine/aot/mod.rs
xxuejie: Sometimes value is not an immediate, it's a register slot needing a runtime value, so we have to do the check at runtime anyway, I think we can skip the check here and just obey the program behavior

src/machine/aot/mod.rs
xxuejie: I thought about it but it will make the generated assembly hard to write. Hence I'm sticking to the simple way here. In fact I think if we really want to optimize this, we can fix the VM so we can skip this. But that will be a bigger change and requires changing our internal bytecode, so I want to leave it till another time.

Another point is this is the same behavior as our assembly interpreter.

src/machine/aot/emitter.rs
```diff
@@ -0,0 +1,443 @@
...
+        let address_value = self.emit_value(address)?;
+        let value_value = self.emit_value(value)?;
+        check_aot_result(unsafe {
+            aot_memory_write(self.aot, address_value, value_value, u32::from(size))
+        })?;
+        Ok(())
+    }
+
+    fn emit_pc_write(&mut self, value: &Value) -> Result<(), Error> {
+        let saved = self.allocator.save();
```
doitian: If it is not considered as a big overhead, I recommend using scope guard to restore the temporary register allocator.

xxuejie: I could be wrong, but based on my understanding, scope guard works in a function, while here we have to work across function boundaries(like [here](https://github.com/nervosnetwork/ckb-vm/blob/05844f3e225700def35817331ba942a6d4c57031/src/machine/aot/emitter.rs#L315-L316), line 316 might choose to emit more values), is that possible with scope guard?

src/machine/aot/emitter.rs
```diff
@@ -0,0 +1,443 @@
...
+        match value {
+            Value::Register(reg) => {
```
doitian: The first two branches are not necessary since `emit_value` in the `_` branch already handled register and immediate.

src/machine/aot/emitter.rs
```diff
@@ -0,0 +1,443 @@
...
+            Write::Register { index, value } => self.emit_register_write(*index, value),
+            Write::Pc { value } => self.emit_pc_write(value),
+            Write::Ecall => check_aot_result(unsafe { aot_ecall(self.aot) }),
+            Write::Ebreak => check_aot_result(unsafe { aot_ebreak(self.aot) }),
+        }?;
+        self.allocator.clear();
+        Ok(())
+    }
+
+    fn emit_memory_write(&mut self, address: &Value, size: u8, value: &Value) -> Result<(), Error> {
```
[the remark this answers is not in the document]
doitian: `clear`ed in `emit`

src/machine/aot/emitter.rs
xxuejie: I see this idea now, thanks. However that would trigger a double mutable borrow here: the call to save is the first borrow, and there will be another `self.emit_value in the omitted` call which triggers a second borrow. We 
 can work around this by storing the allocator value outside of Emitter but that means we will have to pass it around, which is tedious.

doitian: Yes, you are right.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=af949e1bb062789e618606b16b3b7205

xxuejie:
> I think we should run the AOT'd code against RISC-V test suites, is this already set up?

Yes I've actually written the test suite code out and have it all passed locally, but I haven't pushed it to the test suite repo since as soon as I pushed it, it would invalidate all other PRs in this repo. So the idea is once we merge this PR, I will add the missing test suite related code, then we should be all good. Thoughts?

doitian:
> > I think we should run the AOT'd code against RISC-V test suites, is this already set up?
> 
> Yes I've actually written the test suite code out and have it all passed locally, but I haven't pushed it to the test suite repo since as soon as I pushed it, it would invalidate all other PRs in this repo. So the idea is once we merge this PR, I will add the missing test suite related code, then we should be all good. Thoughts?

lgtm

PR #75, opened 2019-06-18
test: avoid make-cov run test two times

u2:


Makefile
```diff
@@ -8,8 +8,8 @@ test-all-features:
 # would skip it.
 cov:
 	cargo clean
-	cargo test --all --features=asm -- --nocapture
-	for file in `find target/debug/ -maxdepth 1 -executable -type f`; do mkdir -p "target/cov/$$(basename $$file)"; kcov --exclude-pattern=/.cargo,/usr/lib,tests "target/cov/$$(basename $$file)" "$$file"; done
+	cargo test --no-run --all --features=asm -- --nocapture
```
doitian: If we just want to build tests only, we can use `cargo build --tests`

u2: I think they are the same.

xxuejie: I vote for `cargo build --tests`, yes the functionalities are the same, but `cargo build --tests` states what we are achieving here better.

PR #78, opened 2019-07-20
fix: Various places where panic might happen in case of malicious data

xxuejie:


src/bits.rs
```diff
@@ -2,7 +2,8 @@
 pub fn roundup(x: u64, round: u64) -> u64 {
     debug_assert!(round.is_power_of_two());
     // x + (((!x) + 1) & (round - 1))
-    x + ((!x).wrapping_add(1) & (round.wrapping_sub(1)))
+    x.overflowing_add((!x).wrapping_add(1) & (round.wrapping_sub(1)))
```
doitian: If we only need the wrapped value, why not just use `wrapping_*` faimly?

PR #79, opened 2019-07-23
chore: update rust to 1.36.0

u2:


rust-toolchain
```diff
@@ -0,0 +1 @@
+1.36.0
```
xxuejie: There was discussion on this before, and the conclusion is we don't want a toolchain file since this is just a library, not a final binary

u2: update

PR #80, opened 2019-08-02
fix: ELF parsing panics in AOT mode

xxuejie:


tests/test_aot.rs
```diff
@@ -267,3 +267,25 @@ pub fn test_aot_wxorx_crash_64() {
...
 }
+
+#[test]
+pub fn test_aot_load_elf_section_crash_64() {
+    let mut file = File::open("tests/programs/load_elf_section_crash_64").unwrap();
+    let mut buffer = Vec::new();
+    file.read_to_end(&mut buffer).unwrap();
+    let buffer: Bytes = buffer.into();
+
+    let result = AotCompilingMachine::load(&buffer.clone(), None);
```
zhangsoledad: probably copy-and-paste and not intended clone here?

xxuejie: Oops, copy-paste issue here

PR #93, opened 2019-12-18
feat: Add step mode for AsmMachine

xxuejie:
The newly added method allows running one instruction in ASM mode, aiding debugging and other usages. For production workload this won't change anything.

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -1,5 +1,5 @@
 use crate::{
-    decoder::build_imac_decoder,
+    decoder::{build_imac_decoder, Decoder},
     instructions::{
         blank_instruction, extract_opcode, instruction_length, is_basic_block_end_instruction,
     },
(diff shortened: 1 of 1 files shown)


src/machine/asm/mod.rs
```diff
@@ -325,6 +325,49 @@ impl<'a> AsmMachine<'a> {
...
+        match result {
+            RET_DECODE_TRACE => (),
+            RET_ECALL => self.machine.ecall()?,
+            RET_EBREAK => self.machine.ebreak()?,
+            RET_MAX_CYCLES_EXCEEDED => return Err(Error::InvalidCycles),
+            RET_OUT_OF_BOUND => return Err(Error::OutOfBound),
+            RET_INVALID_PERMISSION => return Err(Error::InvalidPermission),
+            _ => return Err(Error::Asm(result)),
+        }
+        self.machine.inner_mut().traces[slot] = Trace::default();
```
doitian, line 368: Why reset only on success?

xxuejie: It ensures that every `ckb_vm_x64_execute` only executes one instruction. If we do not reset that, it might happen like this:

* We parse instruction 0x10000, execute it and keep it there
* Later we might run into an instruction at 0x10100 that jumps to 0x10000, if the instruction at 0x10000 is not cleared, `ckb_vm_x64_execute` would first execute the jump instruction at 0x10100, then the normal instruction at 0x10000, which is not the desired behavior we want.

doitian (approved):

jjyr (approved):

PR #96, opened 2020-02-07
AOT mode changes

### LINKED ISSUE #91: AOT mode should be able to share memory bound checking code
Right now we are duplicating all the bound checking code in memory reads/writes in AOT mode, resulting a very large generated binary. However, since AOT works differently from the ASM interpreter, this doesn't have to be the case: ASM interpreter only has limited places where memory checking is needed(bounded by the number of actual instruction types), so duplicating them helps with caches. AOT mode, on the other hands, generates a new series of bound checking logic for every load/store operations, given a non-trivial app, this could result in a very bloated binary.

We could generate a common memory bound checking code, and reuse that in AOT mode, this way we can have a much smaller AOT binary. Since AOT mode just flattens all the original code, CPU branch predictor should have no problems figuring out the right path to jump to.

xxuejie:
This PR made the following changes:

* feat: Share memory write checking code in AOT mode(fixes #91)
* fix: Bound checking bug in AOT mode

Note that AOT mode is not yet used in CKB, we can fix the bug in AOT mode directly. ASM mode will have to wait till the next hardfork.


src/machine/aot/aot.x64.c
```diff
@@ -280,6 +280,53 @@ void aot_finalize(AotContext* context)
 int aot_link(AotContext* context, size_t *szp)
 {
   dasm_State** Dst = &context->d;
+  |->check_write:
```
doitian, line 297: This procedure does not follow the calling convention, I think it's better to add some comment here that what parameters are expected stored in registers, and what is the return value and how to find it.

xxuejie: Already added.

doitian (approved):

driftluo (approved):

Issue #98, opened 2020-02-25
Argv list not endswith NULL in stack initialization


mohanson:
I found that ckb-vm's argv has not endswith null, and then I did a simple test.

The C Standard 5.1.2.2.1/2 says:

> argv[argc] shall be a null pointer.

so create a c program:

```c
#include <stddef.h>

int main(int argc,  char *argv[]) {
	if (argv[argc] == NULL) {
		return 0;
	}
	return 1;
}
```

When built by gcc, and run on Linux, it returns "0".

When built by riscv-unknown-elf-gcc, run on ckb-vm, it returns "1".

You might add a NULL here: https://github.com/nervosnetwork/ckb-vm/blob/develop/src/machine/mod.rs#L160

xxuejie:
We could make this change but has to do it in a non-breaking way, I will give it a thought on how we can achieve this, thanks for the report

xxuejie:
Fixed in #99

Issue #97, opened 2020-02-25
Stack pointer not aligned


mohanson:
During stack initialization, as follows

https://github.com/nervosnetwork/ckb-vm/blob/develop/src/machine/mod.rs#L159

And as riscv-spec-v2.2.pdf, Page 77 says, "In the standard RISC-V calling convention, the stack pointer sp is always 16-byte aligned."

inconsistent behavior here.

xxuejie:
Are you sure you are using the latest spec? I checked the latest version: https://github.com/riscv/riscv-isa-manual/releases/tag/draft-20200213-d1811f4 and I couldn't find the alignment rule as you mentioned

mohanson:
Well, the version I see is indeed older, but I also found it at official website at:

https://riscv.org/specifications/isa-spec-pdf/, Page 125, foot of "C.ADDI16SP"

xxuejie:
I see this now, yes this is an inconsistent behavior, like #98 I think we will have to make it an optional behavior for now and only enables them in the next fork

xxuejie:
Fixed in #99

PR #99, opened 2020-02-26
Introduce version and fix various bugs

### LINKED ISSUE #92: `JALR` causes wrong behavior on AsmMachine when `rs1` and `rd` use the same register.
**Describe the bug**

The following instruction will cause different behavior on DefaultMachine(rust) and AsmMachine(asm). 

``` asm
   1021a:	00000097          	auipc	ra,0x0
   1021e:	fd4080e7          	jalr	-44(ra) # 0x101ee
```

Let's focus on the instruction `1021e`. We expect this instruction to jump to `rs1`(which is `ra`) - `14` and set the `pc + 4` to the `rd`(which is `ra`) register. The DefaultMachine handles this as expected. But unfortunately, the AsmMachine set the registers in a wrong ordering:

https://github.com/nervosnetwork/ckb-vm/blob/develop/src/machine/asm/execute.S#L441

AsmMachine set `rd` first, then read from `rs1`, when the `rs1` and `rd` use the different registers this behavior is correct, but if the `rs1` and `rd` use the same register then the `WRITE_RD` operation overwrite the value of `rs1`.

**How to fix**

To simply fix this problem, we can move the `WRITE_RD` to the after of the  `REGISTER_ADDRESS(RS1)`.

The change may cause the network consensus split. We need to wait for a hard-fork to apply it.

xxuejie:
Since one major use case of CKB VM is on blockchains, we cannot just
change the code in case of bugs. We have to wait for the next fork
event, only then can we upgrade the code. Hence we are introducing a
concept of versions in the VM, whenever we fix a bug that might alter
behavior, we will insert checking code to only include the bug fix in
the next VM version. This way blockchains can expect consistent VM
behavior when they lock specific VM version.

This PR also provides VM version 1 which fixes #92 #97 and #98


definitions/src/instructions.rs
```diff
@@ -24,141 +24,250 @@ pub type InstructionOpcode = u8;
...
+pub const OP_SRAIW: InstructionOpcode = 56;
+pub const OP_SRAW: InstructionOpcode = 57;
+pub const OP_SRL: InstructionOpcode = 58;
+pub const OP_SRLI: InstructionOpcode = 59;
+pub const OP_SRLIW: InstructionOpcode = 60;
+pub const OP_SRLW: InstructionOpcode = 61;
+pub const OP_SUB: InstructionOpcode = 62;
+pub const OP_SUBW: InstructionOpcode = 63;
+pub const OP_SW: InstructionOpcode = 64;
+pub const OP_VERSION1_JALR: InstructionOpcode = 65;
```
doitian, line 90: Any reason to allocate different instructon opcode? From the code base, it seems that `OP_VERSION1_JALR` and `OP_JALR` lead to identical logic.

xxuejie: The 2 opcodes actually do use different logic:

https://github.com/nervosnetwork/ckb-vm/blob/01cf746f32af6be7be2b574b14fe660e63df4719/src/machine/asm/execute.S#L448-L456
https://github.com/nervosnetwork/ckb-vm/blob/01cf746f32af6be7be2b574b14fe660e63df4719/src/machine/asm/execute.S#L1093-L1101

Notice the first 4 lines in each version, they are in different orders. The idea is that version 0 can use the original `OP_JALR`, while version 1 and later uses `OP_VERSION1_JALR`

src/instructions/common.rs
```diff
@@ -80,13 +81,26 @@ pub fn addiw<Mac: Machine>(
 // =======================
 // #  LOAD instructions  #
 // =======================
+fn check_load_boundary<R: Register>(version0: bool, address: &R, bytes: u64) -> Result<(), Error> {
+    if version0 {
+        let address = address.to_u64();
+        let end = address.checked_add(bytes).ok_or(Error::OutOfBound)?;
+        if end == RISCV_MAX_MEMORY as u64 {
```
mohanson, line 88: I am curious why it is not `end >= RISCV_MAX_MEMORY`

xxuejie: The `>` case will be covered by normal memory module bound checking code. This bug only occurs with the `==` case.

mohanson:
If we use `develop` and `release` branches to ensure that the mainnet is not broken, do we still need VERSION0 and VERSION1?

xxuejie:
> If we use `develop` and `release` branches to ensure that the mainnet is not broken, do we still need VERSION0 and VERSION1?

I think we do, based on my understanding of forks, assuming we do a fork update at block 3000000, we will need to use VERSION0 to validate blocks under 3000000 to preserve compatibility, and use VERSION1 to validate blocks after 3000000. That means we will need the concept of version even if we split a release branch.

doitian:
> > If we use `develop` and `release` branches to ensure that the mainnet is not broken, do we still need VERSION0 and VERSION1?
> 
> I think we do, based on my understanding of forks, assuming we do a fork update at block 3000000, we will need to use VERSION0 to validate blocks under 3000000 to preserve compatibility, and use VERSION1 to validate blocks after 3000000. That means we will need the concept of version even if we split a release branch.

It’s likely that we’ll run v0 for codes stored in cells created before a height, and v1 afterwards. Genesis system scripts are exceptions.

mohanson (approved):
LGTM

xxuejie:
> It’s likely that we’ll run v0 for codes stored in cells created before a height, and v1 afterwards. Genesis system scripts are exceptions.

This makes sense but I don't think we need to make exceptions for genesis scripts. They can be treated as scripts created before the fork, hence v0 will be used. Later if there is the need to upgrade them, they might be created into new cells, which can leverage v1 or newer versions.

PR #108, opened 2020-09-16
B extension

mohanson:
Implement the B extension in the rust interpreter, asm machine and aot machine.

<del>Note that aot machine is not yet supported, also, there are two instructions not implemented on the asm machine: `bmatxor` and `bmator`.</del>


src/decoder.rs
```diff
@@ -86,5 +86,6 @@ pub fn build_imac_decoder<R: Register>(version: u32) -> Decoder {
     decoder.add_instruction_factory(rvc::factory::<R>);
     decoder.add_instruction_factory(i::factory::<R>);
     decoder.add_instruction_factory(m::factory::<R>);
+    decoder.add_instruction_factory(b::factory::<R>);
```
xxuejie: Note the name here is `build_imac_decoder`, we should not introduce more extensions here. B should only be activated when someone explicitly specify it now.

mohanson: Understand. I am going to add an additional flag when the machine is initializing, might look like the following `Machine(version=VERSION0, extension=FLAG_IMAC | FLAG_B)`

mohanson: I submitted an update, but CI is broken until this PR merged: https://github.com/nervosnetwork/ckb-vm-test-suite/pull/2

definitions/src/lib.rs
```diff
@@ -16,3 +16,6 @@ pub const MEMORY_FRAME_SHIFTS: usize = 18;
 pub const MEMORY_FRAMESIZE: usize = 1 << MEMORY_FRAME_SHIFTS;
 pub const MEMORY_FRAMES: usize = RISCV_MAX_MEMORY / MEMORY_FRAMESIZE;
 pub const MEMORY_FRAME_PAGE_SHIFTS: usize = MEMORY_FRAME_SHIFTS - RISCV_PAGE_SHIFTS;
+
+pub const ISA_IMAC: u8 = 0b0000_0000;
```
xxuejie: Let's be precise here, the model we actually implemented is `RV64IMC`, not `IMAC`

src/instructions/b.rs
```diff
@@ -0,0 +1,352 @@
+// RISC-V Bitmanip (Bit Manipulation) Extension
+// https://github.com/riscv/riscv-bitmanip
```
xxuejie: Let's document the exact draft version we use now.

definitions/src/lib.rs
mohanson: I will correct it.

xxuejie (dismissed):
I've gone through most of the code here, it is in general well designed and implemented. However, I don't have enough time to carefully analyze the implementation of each individual instructions.

Considering we also have a instruction redesign work depending on this PR. I would recommend a different strategy: we merge this PR first, and apply more testing to make sure our B implementation stays correct.

src/instructions/b.rs
```diff
@@ -0,0 +1,352 @@
...
+                    (0b_001, 0b_11001) => Some(insts::OP_CRC32CH),
+                    (0b_001, 0b_11010) => Some(insts::OP_CRC32CW),
+                    (0b_001, 0b_11011) => Some(insts::OP_CRC32CD),
+                    _ => None,
+                };
+                if inst_opt.is_some() {
+                    return inst_opt.map(|inst| {
+                        Rtype::new(inst, rd(instruction_bits), rs1(instruction_bits), rs2_value).0
+                    });
+                }
```
quake: avoids duplicate pattern matching
```suggestion
                if let Some(inst) = inst_opt {
                    return Some(
                        Rtype::new(inst, rd(instruction_bits), rs1(instruction_bits), rs2_value).0,
                    );
                }
```

brson:
@mohanson I'm seeing a segfault on this branch 95d112c421b6e2f6cd6b3e1c0c84523b1bec8818 in the assembly interpreter

```
$ cargo test --features=asm --test test_b_extension -- --test-threads=1 fsri
    Finished test [unoptimized + debuginfo] target(s) in 0.05s
     Running target/debug/deps/test_b_extension-e6bd6fc635daa745

running 1 test
test test_fsri_decode_bug ... error: test failed, to rerun pass '--test test_b_extension'

Caused by:
  process didn't exit successfully: /home/brian/nervos/ckb-vm/target/debug/deps/test_b_extension-e6bd6fc635daa745 --test-threads=1 fsri (signal: 11, SIGSEGV: invalid memory reference)
```

I haven't been able to debug it yet, but it manifests as freeing an unallocated box

running on wsl

i will keep looking at it another day

brson:
I only see the segfault with rust 1.49.0, not 1.50.0

xxuejie:
Any chance we can check with Rust 1.46.0? This is the version of Rust CKB is now [using](https://github.com/nervosnetwork/ckb/blob/develop/rust-toolchain).

mohanson:
It works well on 1.46, but it is always a serious problem.

mohanson:
Reproduce on MacOS with rust 1.46. It should be a bug.

src/instructions/ast.rs
```diff
@@ -260,6 +274,64 @@ impl Register for Value {
...
+    fn ctz(&self) -> Value {
+        Value::Op1(ActionOp1::Ctz, Rc::new(self.clone()))
+    }
+
+    fn pcnt(&self) -> Value {
+        Value::Op1(ActionOp1::Pcnt, Rc::new(self.clone()))
+    }
+
+    fn rol(&self, rhs: &Value) -> Value {
+        if let (Value::Imm(imm1), Value::Imm(imm2)) = (&self, &rhs) {
```
quake: unnecessary & operator
```suggestion
        if let (Value::Imm(imm1), Value::Imm(imm2)) = (self, rhs) {
```

src/instructions/ast.rs
```diff
@@ -260,6 +274,64 @@ impl Register for Value {
...
+            return Value::Imm(imm1.rotate_left(*imm2 as u32));
+        }
+        Value::Op2(ActionOp2::Rol, Rc::new(self.clone()), Rc::new(rhs.clone()))
+    }
+
+    fn ror(&self, rhs: &Value) -> Value {
+        if let (Value::Imm(imm1), Value::Imm(imm2)) = (&self, &rhs) {
```
quake:
```suggestion
        if let (Value::Imm(imm1), Value::Imm(imm2)) = (self, rhs) {
```

src/instructions/ast.rs
```diff
@@ -260,6 +274,64 @@ impl Register for Value {
...
+            return Value::Imm(imm1.rotate_right(*imm2 as u32));
+        }
+        Value::Op2(ActionOp2::Ror, Rc::new(self.clone()), Rc::new(rhs.clone()))
+    }
+
+    fn slo(&self, rhs: &Value) -> Value {
+        if let (Value::Imm(imm1), Value::Imm(imm2)) = (&self, &rhs) {
```
quake:
```suggestion
        if let (Value::Imm(imm1), Value::Imm(imm2)) = (self, rhs) {
```

src/instructions/ast.rs
```diff
@@ -260,6 +274,64 @@ impl Register for Value {
...
+            return Value::Imm(!((!*imm1).shl(*imm2 as u32)));
+        }
+        Value::Op2(ActionOp2::Slo, Rc::new(self.clone()), Rc::new(rhs.clone()))
+    }
+
+    fn sro(&self, rhs: &Value) -> Value {
+        if let (Value::Imm(imm1), Value::Imm(imm2)) = (&self, &rhs) {
```
quake:
```suggestion
        if let (Value::Imm(imm1), Value::Imm(imm2)) = (self, rhs) {
```

src/instructions/execute.rs
```diff
@@ -2,11 +2,15 @@ use super::{
     super::{machine::Machine, registers::SP, Error},
     common, extract_opcode, instruction_length,
     utils::update_register,
-    Instruction, Itype, Register, Rtype, Stype, Utype,
+    Instruction, Itype, R4type, Register, Rtype, Stype, Utype,
 };
 use ckb_vm_definitions::instructions as insts;
 
-pub fn execute<Mac: Machine>(inst: Instruction, machine: &mut Mac) -> Result<(), Error> {
+#[allow(clippy::cognitive_complexity)]
```
quake: unnecessary lint disable rule?

src/instructions/execute.rs
```diff
@@ -728,16 +732,1041 @@ pub fn execute<Mac: Machine>(inst: Instruction, machine: &mut Mac) -> Result<(),
...
+            None
+        }
+        insts::OP_UNLOADED | insts::OP_CUSTOM_TRACE_END..=255 => {
+            return Err(Error::InvalidOp(op as u8))
+        }
+    };
+    Ok(next_pc)
+}
+
+#[allow(clippy::cognitive_complexity)]
```
quake: unnecessary lint disable rule?

mohanson: I submitted an update that fixes all the view you mentioned https://github.com/nervosnetwork/ckb-vm/pull/108/commits/e3333ca29c705e39289f9bd3f876eb272ef2a1f2

xxuejie (dismissed):

xxuejie (approved):

PR #114, opened 2020-10-09
Update badges

doitian:
* Remove deps.rs badge because it "Failed to analyze repository. The repository you requested might be structured in an uncommon way that is not yet supported."


doitian:
The azure pipeline badge shows "Never built"

xxuejie (approved):

PR #118, opened 2020-11-04
feat: chaos memory mode

mohanson:
- Add a member named `chaos_mode` in `AsmCoreMachine`. When it not equals to zero, It generates random bytes by calling `rand() % 256` in a loop on memory initializing.
- Add a new feature `enable_chaos_mode_by_default` in Cargo.toml, which can modify the `AsmCoreMachine.chaos_mode`'s default value(from 0 to 1).
- Fix truncation of argv pointers.


Cargo.toml
```diff
@@ -21,6 +25,7 @@ goblin = "0.2.0"
 scroll = "0.10"
 ckb-vm-definitions = { path = "definitions", version = "0.19.1" }
 derive_more = "0.99.2"
+rand = "0.7.3"
```
driftluo, line 25: This can be an optional dependency

mohanson: Got!

Cargo.toml
```diff
@@ -13,6 +13,7 @@ default = []
 asm = []
 # Detect if requirements are met, and enable asm feature when we can.
 detect-asm = []
+chaos_mode = ["ckb-vm-definitions/chaos_mode"]
```
xxuejie: Maybe rename this as `enable_chaos_mode_by_default`? It's longer but it tells what it really is.

mohanson: Agree, it makes it clearer

xxuejie:
> It generates random bytes by calling rand() % 256 in a loop on memory initializing.

Notice rand generates a 64-bit random integer, I was curious: why do we truncate it to 8 bits? If we are resetting 8 bytes at a time, maybe we can do the initialization much faster?

mohanson:
> > It generates random bytes by calling rand() % 256 in a loop on memory initializing.
> 
> Notice rand generates a 64-bit random integer, I was curious: why do we truncate it to 8 bits? If we are resetting 8 bytes at a time, maybe we can do the initialization much faster?

We can only ensure that RAND_MAX >= 0x7fff. 

> In GCC (4.6) RAND_MAX = 0x7FFFFFFF (31bit) 
> In MS Visual Studio (2012) RAND_MAX = 0x7FFF (15bit)

Also see: http://www.cplusplus.com/reference/cstdlib/RAND_MAX/

With this in mind, and I think performance is not important for chaos mode also.

xxuejie:
> > > It generates random bytes by calling rand() % 256 in a loop on memory initializing.
> > 
> > 
> > Notice rand generates a 64-bit random integer, I was curious: why do we truncate it to 8 bits? If we are resetting 8 bytes at a time, maybe we can do the initialization much faster?
> 
> We can only ensure that RAND_MAX >= 0x7fff.
> 
> > In GCC (4.6) RAND_MAX = 0x7FFFFFFF (31bit)
> > In MS Visual Studio (2012) RAND_MAX = 0x7FFF (15bit)
> 
> Also see: http://www.cplusplus.com/reference/cstdlib/RAND_MAX/
> 
> With this in mind, and I think performance is not important for chaos mode also.

Gotcha, that makes sense.

src/machine/mod.rs
```diff
@@ -196,8 +196,11 @@ pub trait SupportMachine: CoreMachine {
         for value in values.iter().rev() {
             let address =
                 self.registers()[SP].overflowing_sub(&Self::REG::from_u8(Self::REG::BITS / 8));
-
-            self.memory_mut().store32(&address, value)?;
+            if Self::REG::BITS == 64 {
```
xxuejie: Can we enable this only in VERSION1, so as to preserve compatibility?

mohanson: Got!

xxuejie (approved):

src/machine/asm/execute.S
```diff
@@ -177,6 +177,40 @@
...
+  jmp 5b; \
+6: \
+  POSTCALL;
+
+/*
+ * Make sure TEMP1 stores the frame index.
+ */
+#define INITED_MEMORY \
+  movzbl CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_CHAOS_MODE(MACHINE), TEMP2d; \
+  cmp $0, TEMP2d; \
```
jjyr, line 206: Since we already use the rust compiling conditions, we should be able to remove this branch and determine whether to enable or disable the chaos mode in GCC preprocessor.

doitian (approved):

src/machine/asm/execute.S
xxuejie: Preprocessor is a compile-time thing, the other feature I think we can enable, is that a user can choose to enable or disable chaos mode at runtime via APIs. The compiling condition merely affects default behavior here.

PR #124, opened 2020-12-10
feat: support to modify the chaos_seed

mohanson:
Support to generate fixed random data under `chaos_mode` by setting `chaos_seed`.

- On windows: use `srand(seed)`+ `rand()`.
- On posix: use `rand_r(seedp)` since `srand` + `rand` is not thread-safe on posix.


src/machine/asm/execute.S
```diff
@@ -177,26 +179,54 @@
   CALL_MEMSET; \
   POSTCALL;
 
+#ifdef IS_WINDOWS
 #define RANDOM_MEMORY \
   PREPCALL; \
+  mov CHAOS_SEED, ARG1; \
+  call srand; \
```
xxuejie: I think we might only need to call `srand` once when we initialize the VM, right now the implementation calls srand before each page initialization, this would result in the fact that each page is filled with exactly the same data, correct?

mohanson: The seed used for the second `srand()` is the last data generated by the first `srand()` + `rand()`

xxuejie: I see, in that sense this works.

xxuejie (dismissed):

src/machine/aot/aot.x64.c
```diff
@@ -344,55 +346,65 @@ int aot_link(AotContext* context, size_t *szp)
...
     | jmp <1
     |2:
     | postcall
     | ret
   |.endif
   |->inited_memory:
-  | lea rdx, machine->chaos_mode
-  | mov dl, byte [rdx]
-  | cmp dl, 0
+  | lea r8, machine->chaos_mode
```
xxuejie: Any particular reason for this register change?

mohanson: rdi stores the address of the machine, if I change it at this time, then in the subsequent code, `machine->chaos_seed` will returns unexpected address

xxuejie: Maybe there are some other registers, like eax, ecx or edx we can use? The problem with r8/r8d is that they are x64 registers which will require longer instruction coding, hence longer code size. I know, we are only talking 1 or 2 bytes here, but it will just be a tiny change if we do have a register available around.

mohanson: I made a mistake here, the machine exists in `rdi` not `rdx`, I think the modification here is make no sense, but i can't run test without this modifcation, there must be a bug elsewhere.

xxuejie (dismissed):

xxuejie (approved):

Issue #109, opened 2020-09-17
Suspend/resume a running VM


xxuejie:
We should be able to suspend a running CKB VM, save the states somewhere, and resume the previously running VM at a latter time, or even on a different machine. One potential design here might be:

* A signal is sent from another thread to CKB VM to start suspending
* CKB VM will pack the current register states, as well as any dirty memory pages in a packed data structure. Notice we can easily exclude the following memory pages from dirty memory page list:
    + Page with all zeros as the data
    + All executable pages, since they must come from some binaries
    + Writable pages coming from an ELF, but hasn't been modified in any way.
* Then the packed data structure can be restored elsewhere into a running VM instance that is exactly the same as what it is before suspending. Execution can then continue from there.

mohanson:
In the above design, the ELF binary needs to be reloaded again when the VM is resumed. Have you considered just dumping the entire memory? there may be better performance.

xxuejie:
I was under the impression that the ELF binary will be available elsewhere(for example: in CKB this will be in a cell). Dumping the entire memory might result in much bigger memory consumption. Consider the fact that it is typically long running programs that would require suspend/resume, I would bet that loading the binary again is a acceptable tradeoff to reduce load on memory & (possibly) disk requirements.

xxuejie:
One additional idea from @doitian: instead of using a signal, we can simply allow suspending when the available cycles are all consumed. That will suit CKB's use case much better.

PR #125, opened 2020-12-23
feat: suspend/resume the vm

mohanson:
For #109


src/machine/aot/aot.x64.c
```diff
@@ -369,6 +369,14 @@ int aot_link(AotContext* context, size_t *szp)
   | and edx, CKB_VM_ASM_MEMORY_FLAG_WXORX_BIT
   | cmp edx, CKB_VM_ASM_MEMORY_FLAG_WRITABLE
   | jne >4
+  /*
+   * Set the page as dirty
+   */
+  | lea rdx, machine->flags
+  | movzx edx, byte [rdx+rcx]
+  | and edx, CKB_VM_ASM_MEMORY_FLAG_DIRTY
```
xxuejie: I think this should be an or, correct?

src/machine/aot/aot.x64.c
```diff
@@ -369,6 +369,14 @@ int aot_link(AotContext* context, size_t *szp)
...
+  | lea rdx, machine->flags
```
xxuejie: This line already changes the value of rdx, which means `dl` below is not a correct value.

src/snapshot.rs
```diff
@@ -0,0 +1,76 @@
...
+    snap.pc = machine.pc().to_u64();
+
+    for i in 0..RISCV_PAGES {
+        let flag = machine.memory_mut().fetch_flag(i as u64)?;
+        if flag & FLAG_DIRTY != 0 {
+            let addr_from = i << RISCV_PAGE_SHIFTS;
+            let addr_to = (i + 1) << RISCV_PAGE_SHIFTS;
+
+            let mut page = vec![0; RISCV_PAGESIZE];
+            for i in (addr_from..addr_to).step_by(8) {
```
xxuejie, line 54: Can we add some comments to document how snapshot works?

src/machine/aot/aot.x64.c
mohanson: Yes, I wrote the wrong logic.

src/snapshot.rs
mohanson: I submitted an update, and now the comments will explain what we did.

```rs
// Snapshot provides a mechanism for suspending and resuming a virtual machine.
//
// When cycle limit is too low, our work won't be finished when we hit it and
// a "max cycles exceeded" error will be returned. At this time, a snapshot could
// be created, we can create a new virtual machine from this snapshot, provide
// more cycles and then continue to run it.
//
// For the following data, we simply save them, and then restore them.
//   - machine.version
//   - machine.pc
//   - machine.registers
//
// For memory, the situation becomes more complicated. Every memory page has
// page flag where each page flag stores a optional FLAG_DIRTY. When this page
// is wrote, the memory instance will set this bit to 1, this helps us keep
// track of those pages that have been modified. After `machine.load_elf`, We
// clean up all dirty flags, so after the program terminates, all pages marked
// as dirty are the pages that have been modified by the program. We only store
// these pages in the snapshot.
```

xxuejie (approved):

PR #126, opened 2020-12-30
feat: allow analysis elf before initialization vm

mohanson:
Add a new package level function `parse_elf()` and  a new method `DefaultMachine.load_program_elf(&mut self, program: &Bytes, args: &[Bytes], elf: &Elf,)`, this allows analysis of the elf file before initializing the vm, while ensuring that it only needs to be parsed once.


xxuejie:
Can we put this on hold for now? I want to spend some time to think the best way to approach this feature.

mohanson:
Okay, we should treat it cautiously.

xxuejie (approved):

xxuejie:
The PR title needs updating @mohanson

mohanson:
> The PR title needs updating @mohanson

Done.

PR #127, opened 2021-02-23
Give every CI config a name

brson:
Looking at the CI results, it is not easy to see what each job is. This should make it more clear.


brson:
Failure looks unrelated to this change.

PR #130, opened 2021-03-05
chore: bump goblin

mohanson:
Different version of goblin crates will be used for different version of machines:

- when machine.version == VERSION0, goblin = "0.2.0"
- when machine.version == VERSION1, goblin = "0.3.4"

Fix:
- In the test case, AsmMachineCore.version is now fixed to VERSION0 instead of a random value.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,8 @@ enable-chaos-mode-by-default = ["ckb-vm-definitions/enable-chaos-mode-by-default
 [dependencies]
 byteorder = "1"
 bytes = "0.5.4"
-goblin = "0.2.0"
+goblin_v020 = { package = "goblin", version = "0.2.0" }
+goblin_v034 = { package = "goblin", version = "0.3.4" }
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -18,7 +18,8 @@ enable-chaos-mode-by-default = ["ckb-vm-definitions/enable-chaos-mode-by-default
 [dependencies]
 byteorder = "1"
 bytes = "0.5.4"
-goblin = "0.2.0"
+goblin_v0 = { package = "goblin", version = "0.2.0" }
```
xxuejie: I think it might be better if we stick to goblin's own versioning here, such as `goblin_v020` and `goblin_v034`. CKB VM has the concept of v0 and v1, but goblin does not.

mohanson: Agree with this view.

xxuejie (dismissed):

xxuejie (dismissed):

xxuejie (approved):

PR #131, opened 2021-03-10
perf: optimize decode bits fn

quake:
use a faster path to decode instruction bits, compare with develop branch, it improved secp256k1_bench around 23%

this PR:
```
interpret secp256k1_bench via assembly                                                                            
                        time:   [8.8069 ms 8.8639 ms 8.9184 ms]

interpret secp256k1_bench via assembly                                                                            
                        time:   [9.0110 ms 9.0526 ms 9.0955 ms]

interpret secp256k1_bench via assembly                                                                            
                        time:   [9.0318 ms 9.1958 ms 9.3681 ms]

interpret secp256k1_bench via assembly                                                                            
                        time:   [8.7876 ms 8.8759 ms 8.9668 ms]

interpret secp256k1_bench via assembly                                                                            
                        time:   [8.9817 ms 9.0287 ms 9.0777 ms]

avg: 9.00338
```

develop branch:
```
interpret secp256k1_bench via assembly                                                                            
                        time:   [11.763 ms 11.802 ms 11.845 ms]

interpret secp256k1_bench via assembly                                                                            
                        time:   [11.675 ms 11.729 ms 11.787 ms]

interpret secp256k1_bench via assembly                                                                            
                        time:   [11.727 ms 11.783 ms 11.841 ms]

interpret secp256k1_bench via assembly                                                                            
                        time:   [11.342 ms 11.463 ms 11.590 ms]

interpret secp256k1_bench via assembly                                                                            
                        time:   [11.794 ms 11.881 ms 11.999 ms]

avg: 11.7316
```

apply the same optimization on release branch, around 5% improvement:
```
interpret secp256k1_bench via assembly                                                                             
                        time:   [2.2648 ms 2.2709 ms 2.2779 ms]
interpret secp256k1_bench via assembly                                                                             
                        time:   [2.3300 ms 2.3573 ms 2.3880 ms]
interpret secp256k1_bench via assembly                                                                             
                        time:   [2.2791 ms 2.2889 ms 2.3005 ms]
interpret secp256k1_bench via assembly                                                                             
                        time:   [2.2343 ms 2.2375 ms 2.2408 ms]
interpret secp256k1_bench via assembly                                                                             
                        time:   [2.2488 ms 2.2637 ms 2.2807 ms]
interpret secp256k1_bench via assembly                                                                             
                        time:   [2.2491 ms 2.2547 ms 2.2609 ms]

avg: 2.27883333
```

release branch
```
interpret secp256k1_bench via assembly                                                                             
[...]

PR #132, opened 2021-03-10
perf: optimize decode bits fn

quake:
ref: https://github.com/nervosnetwork/ckb-vm/pull/131


quake:
don't apply for release branch

PR #133, opened 2021-03-12
refactor: simplify memory trait design

quake:
This PR replaced `Memory` trait's `Register` generic type by associated type, which simplifed the fn parameter and removed some dynamic dispatches, for example:
https://github.com/nervosnetwork/ckb-vm/blob/3a4484b0472193c29329560c0d461096b4352e17/src/memory/mod.rs#L64

We can move fn `set_dirty` and `check_permission` to `Memory` trait as a default implementation after this PR.

### DIFF
--- a/src/memory/wxorx.rs
+++ b/src/memory/wxorx.rs
@@ -5,29 +5,28 @@ use super::{
 };
 
 use bytes::Bytes;
-use std::marker::PhantomData;
 
-pub struct WXorXMemory<R: Register, M: Memory<R>> {
+pub struct WXorXMemory<M: Memory> {
     inner: M,
-    _inner: PhantomData<R>,
 }
 
-impl<R: Register, M: Memory<R> + Default> Default for WXorXMemory<R, M> {
+impl<M: Memory + Default> Default for WXorXMemory<M> {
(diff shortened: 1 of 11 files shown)


src/memory/wxorx.rs
```diff
@@ -5,29 +5,28 @@ use super::{
 };
 
 use bytes::Bytes;
-use std::marker::PhantomData;
 
-pub struct WXorXMemory<R: Register, M: Memory<R>> {
+pub struct WXorXMemory<M: Memory> {
     inner: M,
-    _inner: PhantomData<R>,
```
mohanson, line 12: I am curious why we add a PhantomData before, and why it makes sense to delete it for now.

quake: when using generics type in `Memory` trait, we have to annotate the types in each implementation.

mohanson (dismissed):

xxuejie (dismissed):

quake:
rebased with develop to resolve conflict, please review again, thanks

mohanson (approved):

xxuejie (approved):

PR #137, opened 2021-03-15
test: rvc instruction in page end

mohanson:
Ref https://github.com/nervosnetwork/ckb-vm/pull/131#pullrequestreview-611271106

Add a new test case:

```text
.section .text
.global _start
_start:
    nop -----------+
    ... 1985 times |
    nop -----------+
    li a2, 1
    li a1, 3
    c.sub a1, a2 <----- 4094 offset of the page
    li a0, 0
    li a7, 93
    ecall
```


quake (dismissed):

xxuejie:
This test case has a slight problem: the page after `c.sub` instruction, is still executable. This way, we won't be able to test again edge cases needed in #131. I think we might want a case like this:

```
# address 100
li a0, 0
li a7, 93
ecall
# ...
# address 4090
li a0, 100
c.jr a0
# starting here, we need to have a writable but not executable page
```

This way we can really have an edge case, where only 2 bytes at the page boundary are executed.

mohanson:
I modified the test, please review it again @xxuejie 

```
.section .text
.global _start
_start:
    li a0, 0x100c4  <----- 0x100b0
    c.jr a0
    nop
    li a0, 0        <----- 0x100ba
    li a7, 93
    ecall
    nop             <----- 0x100c4
    ...
    nop
    li a0, 0x100ba
    c.jr a0         <----- 0x10ffe
.section .data      <----- 0x11000
n0:
    .word 4
n1:
    .word 2

```

xxuejie (approved):

PR #139, opened 2021-03-22
Add a CI config for running under miri

brson:
I said earlier I would add such a configuration. This takes around 4 minutes to run locally. If that is too long for the CI I can try reduce the time further.

cc @mohanson

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,6 +86,14 @@ matrix:
             - git
             - build-essential
       env: SUITE=ci-chaos
+    - name: ci-miri
+      rust: nightly-2021-03-20
+      addons:
+        apt:
+          packages:
+            - git
+            - build-essential
+      env: SUITE=ci-miri
     - name: Test Suite
       rust: 1.46.0
       dist: xenial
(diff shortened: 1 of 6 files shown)


brson:
Looks like the miri job ran correctly.

xxuejie:
One question: is there we can employ some sort of CLI-controllable flags, so instead of using this in the code:

```
#[cfg_attr(miri, ignore)]
```

we can turn the additional tests on and off in CLI part. The result for the change, is that we can run the core and fast tests in CI, but run all the tests locally.

Is something like this possible?

brson:
@xxuejie I've pushed a change that adds a `miri-ci` feature and turned it on for CI. This feature disables slow tests. Without the feature all tests are run except for `test_b_extension`.

I ran `test_b_extension` for 9 hours and it took 13 GB of RAM, and I never saw it finish. I can though leave it enabled if you prefer.

The remaining tests take me about 3.5 hours to run.

For multiple reasons it would be nice to break up the B extension test suite into individual tests, but that is a larger upstream task.

xxuejie (approved):

PR #140, opened 2021-03-23
feat: redesign instruction

mohanson:
- Redesigned the format of instructions
- Inner RVC instruction and Inner VERSION0/VERSION1 instruction has been removed
- Some MOP instructions have been added experimentally

I think we need to discuss which MOP instructions should be added and how to test them, but they are not the focus of this PR.


src/decoder.rs
```diff
@@ -75,24 +80,284 @@ impl Decoder {
...
-    let mut decoder = Decoder::new(version);
+pub fn build_decoder<R: Register>(isa: u8) -> Decoder {
+    let mut decoder = Decoder::new();
     decoder.add_instruction_factory(rvc::factory::<R>);
     decoder.add_instruction_factory(i::factory::<R>);
     decoder.add_instruction_factory(m::factory::<R>);
     if isa & ISA_B != 0 {
         decoder.add_instruction_factory(b::factory::<R>);
     }
+    decoder.mop = isa & ISA_MOP != 0;
```
xxuejie: This feels weird to me, why not make `mop` a parameter of `Decoder::new`?

xxuejie (dismissed):
Elegant work! Just one minor comment here.

src/decoder.rs
mohanson: It's a design defect, fixed now.

quake (dismissed):

xxuejie (approved):

PR #142, opened 2021-03-30
fix: writable page should not be frozen

mohanson:


xxuejie (changes requested):
This is a breaking change, it should also only be enabled in version 1 and beyond.

xxuejie (commented):
Can we add test cases to ensure the correct behavior for both version 0 and 1?

mohanson:
@xxuejie I added the tests

xxuejie (approved):

Issue #143, opened 2021-04-05
Upgrade goblin to 0.3.5


brson:
**Describe the bug**

Goblin <= 0.3.4 contains an arithmetic overflow decoding bad elf metadata.

It is fixed by https://github.com/m4b/goblin/pull/260 but not yet released.

The overflow in question is here https://github.com/m4b/goblin/blob/0dbaa385dbba8f56267febe0ff8fd7eda23acc39/src/elf/mod.rs#L409

In debug mode the loader will panic. In release mode, further sanity checks in the loader should prevent anything catastrophic, but I have only read the code briefly.

**To Reproduce**

I can provide an elf binary that reproduces the error on request.

cc @mohanson @xxuejie

mohanson:
I think we can add a test case first and mark the test case as `#[ignore]` until goblin releases a new version

brson:
I'll add that case.

PR #144, opened 2021-04-06
fix: roll back some changes to elf parsing

mohanson:
In the [previous commit](https://github.com/nervosnetwork/ckb-vm/commit/8b3f65da3ea77243a76c23a386313ff9a397d0c7), full elf parsing was used instead of partial elf parsing, which would trigger some bugs in goblin.

Now this part of the modification is rolled back, so that the logic is roughly the same as the release branch. 

ref: https://github.com/nervosnetwork/ckb-vm/issues/143


xxuejie (dismissed):

xxuejie (dismissed):
Goblin now has 0.4.0 released, we might want to upgrade to that version in this PR. Other than that this is all looking good.

mohanson:
I will update it right now.

xxuejie (approved):

PR #146, opened 2021-04-28
Use soft popcnt

mohanson:
These changes has been fuzzing for several hours.


xxuejie:
Can we add some tests here?

mohanson:
> Can we add some tests here?

Sure

mohanson:
Since b_extension already has [related tests](https://github.com/riscv/riscv-bitmanip/blob/master/tests/test_zbb.c#L13), the new test cases added this time is not so much.

xxuejie (approved):

PR #159, opened 2021-05-21
chore: add take api

driftluo:
The snapshot only records the internal state(registers and memory) of the running VM. To fully restore the scene, we need to support all syscall、cycle calculation function and debugger function, rebuilding is too expensive each time, this PR will be added all `take_*` API to get the information inside the VM to facilitate the reconstruction

### DIFF
--- a/src/machine/mod.rs
+++ b/src/machine/mod.rs
@@ -538,6 +538,18 @@ impl<'a, Inner: SupportMachine> DefaultMachine<'a, Inner> {
         self.inner
     }
 
+    pub fn take_cycle_func(&mut self) -> Option<Box<InstructionCycleFunc>> {
+        self.instruction_cycle_func.take()
+    }
+
+    pub fn take_syscalls(&mut self) -> Vec<Box<dyn Syscalls<Inner> + 'a>> {
+        ::std::mem::take(&mut self.syscalls)
+    }
+
+    pub fn take_debugger(&mut self) -> Option<Box<dyn Debugger<Inner> + 'a>> {
+        self.debugger.take()
+    }
+
     pub fn exit_code(&self) -> i8 {
         self.exit_code
     }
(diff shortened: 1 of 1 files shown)


xxuejie:
I'm not sure about this change. Several considerations:

1. What work does rebuilding the syscalls and all surrounding environment requires? Which part is considered as the expensive part?
2. Have we performed analysis on the memory consumption when packing all the syscall structs together? Snapshotting VM alone can already be a memory consuming task, I wonder what the overall memory requirement would be if all syscalls are also packed.
3. A lot of CKB syscalls we use now has [this structure](https://github.com/nervosnetwork/ckb/blob/3ee17124c905a03450456ca84f946c38090a11e7/script/src/syscalls/load_cell.rs#L19-L25), where we are only take references of the data structure owned by top level TransactionScriptsVerifier. This is the part I wasn't sure: if we take the syscalls out of the VM instance, we might need to deal with lifetime hassles that can bubble up fairly quickly. Maybe there is a solution, I just wasn't sure a clear and simple path exists.

driftluo:
Because of the lifetime, ckb may only be able to use the API `take_cycle_func`, the rest of the syscall may need to be rebuilt, but if the syscall is `‘static`, the existence of these APIs will have positive benefits for rebuilding the overall environment.

What I mean is the step of [loading all cells](https://github.com/nervosnetwork/ckb/blob/develop/script/src/verify.rs#L99), because this operation currently cannot save the context. When using snapshots, it may be necessary to rewrite the state preservation of load cells, but this will bring a new problem. During the execution process, the saved cell state has been changed. It is a wrong behavior to continue execution. This is a place that needs careful consideration. 

In addition, these `take` APIs did not bring any negative overall, I personally think it is worth doing this

xxuejie:
To me `cycle_func` is just a pure function, I wonder how much benefit that's gonna bring.

I totally agree with you that cell loading might result in IO operations that could be a slowdown point. But a balancing factor, is that all those cells would take up memory space, where you are playing the game of over-consuming memory for fewer IO operations. I'm not saying we shouldn't do this, I'm just saying we will probably need numbers to showcase which side we want to lean against.

As for the state changing problem, I'm afraid I lost the context. To me all the cells used by a transaction would stay unchanged no matter when and where we run the scripts in the transaction. This is determined by the whole UTXO model. Even if you discard all the loaded cells, and at a later time, choose to resume and re-load all the cells, you should still be loading the same contents, otherwise we are facing a bigger problem.

Lastly, I believe APIs, especially public facing APIs would shape the way people use the library, so I cannot be on board at the moment with the changes only because they `did not bring any negative overall`

driftluo:
Okay, let me think about the issue here again and close this PR first

Issue #161, opened 2021-05-28
Use `reset()` as a general means to reuse machine


mohanson:
In the current CKB, a new machine needs to be created every time a script is executed, which means a malloc with a size of at least 4M. I consider using `reset()` to reset the state of the machine after each script execution is over, as shown below

```rs
// Start execute first script
machine.run()

// Reset states of machine
machine.reset()
machine.set_max_cycles(_)
machine.set_cycles(0)

// Start execute next script
machine.load_program(_)
machin.run()
```

But now there is a problem to think about: this approach breaks isolation. 

cc @driftluo

xxuejie:
I think this approach works: the design of reset is ensures that after resetting the machine, all data from previous machine has been *erased*, can you elaborate on why this new approach breaks isolation?

xxuejie:
In addition, I wasn't sure on this snippet:

```
machine.reset()
machine.set_max_cycles(_)
machine.set_cycles(0)
```

`reset` is designed to, well, reset everything, so why bother setting cycles to 0 here?

driftluo:
>  why bother setting cycles to 0 here?

When CKB Fork, the temporarily stored VM can have 2 versions. The control of cycles cannot be controlled by the VM. It is necessary to reset the `max_cycles` at the CKB level at each reset.

mohanson:
1) `.cycles()` record all the cycles used by the machine from birth to death, not just a specific script. Therefore `reset()` does not reset cycles, and its behavior conforms to `exec()`.

2) From a low-level perspective, the data recorded in the memory by the previous script has not been erased.

xxuejie:
I see the point of resetting cycles in the term of suspend/resume. But I wasn't sure why the versions of VM affects cycles. Current `reset` function, as far as I know, won't clear syscalls, won't swap to different version. So if we are using different versions of VM, we will have to initialize different VM instances, we cannot use `reset` there.

xxuejie:
> From a low-level perspective, the data recorded in the memory by the previous script has not been erased.

But the VM, after resetting, can never read the data written by previous scripts. Lazy initialization will take place here: once you try to access a page, it will be initialized, and zeroed first. So unless there is a bug, I don't think we have any isolation problem.

driftluo:
[...]

PR #162, opened 2021-05-28
fix: reset() correctly set cycles and max_cycles now

mohanson:
ref https://github.com/nervosnetwork/ckb-vm/issues/161

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -303,7 +303,7 @@ impl SupportMachine for Box<AsmCoreMachine> {
         Some(self.max_cycles)
     }
 
-    fn reset(&mut self) {
+    fn reset(&mut self, max_cycles: Option<u64>) {
         self.registers = [0; RISCV_GENERAL_REGISTER_NUMBER];
         self.pc = 0;
         self.flags = [0; RISCV_PAGES];
@@ -311,6 +311,9 @@ impl SupportMachine for Box<AsmCoreMachine> {
             self.traces[i] = Trace::default();
         }
         self.frames = [0; MEMORY_FRAMES];
+        self.cycles = 0;
+        debug_assert!(max_cycles.is_some());
+        self.max_cycles = max_cycles.unwrap_or(0);
         self.reset_signal = 1;
     }
 
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -312,8 +312,8 @@ impl SupportMachine for Box<AsmCoreMachine> {
         }
         self.frames = [0; MEMORY_FRAMES];
         self.cycles = 0;
-        debug_assert!(max_cycles.is_some());
-        self.max_cycles = max_cycles.unwrap_or(0);
+        // AsmMachine always needs a max_cycles, caller needs to guarantee this.
+        self.max_cycles = max_cycles.unwrap();
         self.reset_signal = 1;
     }
 
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -299,11 +299,11 @@ impl SupportMachine for Box<AsmCoreMachine> {
         self.cycles = cycles;
     }
 
-    fn max_cycles(&self) -> Option<u64> {
-        Some(self.max_cycles)
+    fn max_cycles(&self) -> u64 {
+        self.max_cycles
     }
 
-    fn reset(&mut self, max_cycles: Option<u64>) {
+    fn reset(&mut self, max_cycles: u64) {
         self.registers = [0; RISCV_GENERAL_REGISTER_NUMBER];
         self.pc = 0;
         self.flags = [0; RISCV_PAGES];
@@ -312,9 +312,7 @@ impl SupportMachine for Box<AsmCoreMachine> {
         }
         self.frames = [0; MEMORY_FRAMES];
         self.cycles = 0;
-        if let Some(cycles) = max_cycles {
-            self.max_cycles = cycles;
-        }
+        self.max_cycles = max_cycles;
         self.reset_signal = 1;
     }
 
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -312,8 +312,9 @@ impl SupportMachine for Box<AsmCoreMachine> {
         }
         self.frames = [0; MEMORY_FRAMES];
         self.cycles = 0;
-        // AsmMachine always needs a max_cycles, caller needs to guarantee this.
-        self.max_cycles = max_cycles.unwrap();
+        if let Some(cycles) = max_cycles {
+            self.max_cycles = cycles;
+        }
         self.reset_signal = 1;
     }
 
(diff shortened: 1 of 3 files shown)


src/machine/asm/mod.rs
```diff
@@ -303,14 +303,17 @@ impl SupportMachine for Box<AsmCoreMachine> {
...
         self.registers = [0; RISCV_GENERAL_REGISTER_NUMBER];
         self.pc = 0;
         self.flags = [0; RISCV_PAGES];
         for i in 0..TRACE_SIZE {
             self.traces[i] = Trace::default();
         }
         self.frames = [0; MEMORY_FRAMES];
+        self.cycles = 0;
+        debug_assert!(max_cycles.is_some());
+        self.max_cycles = max_cycles.unwrap_or(0);
```
driftluo: https://github.com/nervosnetwork/ckb-vm/blob/develop/definitions/src/asm.rs#L61

Inconsistent with the default behavior

By the way, why asm machine defaults to u64 max and rust machine defaults to 0

mohanson: 1) The purpose here is to ensure that max_cycles is not None, and unwrap_or(0) can be safely ignored.

2) See: https://github.com/nervosnetwork/ckb-vm/blob/develop/definitions/src/asm.rs#L58-L60

src/machine/asm/mod.rs
```diff
@@ -303,14 +303,17 @@ impl SupportMachine for Box<AsmCoreMachine> {
...
         self.reset_signal = 1;
```
driftluo, line 316: If I read correctly when reset_signal is 1, the aot code will be discarded, so the behavior of aot machine should be problematic

https://github.com/nervosnetwork/ckb-vm/blob/develop/src/machine/asm/mod.rs#L371-L373

mohanson: Yes, in this case, aot will be converted to asm to run. This is because it is not possible to add new aot code for SupportMachine in the exec system call .

driftluo: Ignore exec syscall first.

If it is a normal aot VM, the behavior of losing aot code after reset seems a bit weird. User expects reset to return to the initialization state with aot code?

xxuejie: > This is because it is not possible to add new aot code for SupportMachine in the exec system call .

We should fix this instead. Now that we think about it, it would make more sense to provide AoT code, when we are doing `load_program`, not when we are initializing the VM.

mohanson: In this case, you can't just `reset()` and then `load_program()`, because the aot code is not generated from `load_program`. The correct approach is:

```
machine.reset()

let mut aot_machine = AotCompilingMachine::load(_);
let code = aot_machine.compile().unwrap();
machine.aot_code = code
machine.reset_signal = false;

machine.load_program(_);
machine.run();
```

mohanson: > We should fix this instead. Now that we think about it, it would make more sense to provide AoT code, when we are doing load_program, not when we are initializing the VM.

We can solve this problem in the next PR

src/machine/asm/mod.rs
xxuejie: This is weird to me, I think we should stick to either one:

* ensure users always provide a max cycle in the reset call here
* or fix the machine, so it can have max cycle or not

This `unwrap_or` call here, does not make too much sense to me.

src/machine/asm/mod.rs
xxuejie: A different PR works, but we should definitely fix this.

src/machine/asm/mod.rs
mohanson: Since AsmMachine always needs to set max cycles, I think the option 1) is better.

driftluo: Maybe this solution is better:
```
if let Some(cycles) = max_cycles {
  self.max_cycles = cycles;
}
```

mohanson: Accept.

driftluo (dismissed):

src/machine/asm/mod.rs
xxuejie: I do have a concern regarding this solution: at a lower level, AsmMachine does not accept an option, you should always have a max cycles set, one typical solution, is to use `u64::max_value()` in case one does not require a max cycle. This brings a problem: when we are using None in reset function, what behavior are we achieving? We now have 2 different APIs:

* When you are initializing a VM, you fill in `u64::max_value()` when you do not have a cycle to set
* When you are resetting, you can use `None` or `u64::max_value()`

In this sense, I'm not sure using Option here, is a good solution.

driftluo: I think most users will use the new method to create an asm machine, reset with None means keep the original max cycles

mohanson: Option is necessary because RustMachine requires it. 

https://github.com/nervosnetwork/ckb-vm/blob/develop/src/machine/mod.rs#L274

xxuejie: The root problem here, is that the original max cycles are kept, so we are just introducing a new rule to remember here. The Option here, does not actually provide what a typical Option value provides. In this sense, I would more be willing to remove Option here and always require a value. If a user does not have a proper value to set, they can just use `u64::max_value()`

mohanson: For RustMachine's reset, None means that there is not limit for cycles.

For AsmMachine, None should perform a similar function. So if it is None, is it better to use u64::max_value()? 

```rs
max_cycles.unwrap_or(u64::max_value())
```

xxuejie: I think we should just have consistent behavior.

If we want to support Option, we should support it in both places, when you are initializing a VM, and when you are resetting. If we cannot support it, we should not provide an Option here.

xxuejie: Part of me think it might be better just to ensure you always have a max cycles set, and remove that Option in RustMachine.

If one does not care for cycles, they can always choose not to provide a cost function, or a function that always return 0. That to me will be a better solution.

mohanson: > Part of me think it might be better just to ensure you always have a max cycles set, and remove that Option in RustMachine.

It is a perfect solution.

src/machine/mod.rs
```diff
@@ -75,7 +75,7 @@ pub trait SupportMachine: CoreMachine {
     fn set_running(&mut self, running: bool);
 
     // Erase all the states of the virtual machine.
-    fn reset(&mut self);
+    fn reset(&mut self, max_cycles: Option<u64>);
```
doitian: The interface is confusing. Does `None` mean keeping the original max cycles or clear the max cycles limit?

mohanson: I have removed Option wrapper of max_cycles in the latest commit, please review it again

xxuejie (approved):

driftluo (approved):

PR #164, opened 2021-06-01
feat: support dynamic reset max cycles

driftluo:
#### Why need this?

In the scenario where the program is executed step by step, after the VM executes to the initialization max, jump out of execution to check whether the snapshot operation needs to be executed, if not, continue execution with the current VM state. 

At this time, `set_max_cycles` is required because the remaining executable cycles may not be equal to the initialized one

The code currently available for [reference](https://github.com/driftluo/ckb/blob/tmp-vm/tx-pool/src/chunk_process.rs#L174)

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -303,10 +303,6 @@ impl SupportMachine for Box<AsmCoreMachine> {
         self.max_cycles
     }
 
-    fn set_max_cycles(&mut self, cycles: u64) {
-        self.max_cycles = cycles;
-    }
-
     fn reset(&mut self, max_cycles: u64) {
         self.registers = [0; RISCV_GENERAL_REGISTER_NUMBER];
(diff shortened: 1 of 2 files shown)


mohanson (dismissed):

xxuejie:
I could understand the rationale, but can we remove set_max_cycles from SupportMachine trait? Syscalls at least should not touch them. The outsider code in the reference has direct reference to the full VM instance, adding a method in the top VM instance level should work

driftluo:
> can we remove set_max_cycles from SupportMachine trait?

The latest commit puts the method on `DefaultCoreMachine`, `AsmMachine` can actually access the internal `Box<AsmCoreMachine>`, no need to add methods, but I still added it, so now there is such a method at the top level

mohanson (approved):

xxuejie (approved):

Issue #163, opened 2021-05-31
Remove confusing constructor like `machine::latest()` or `machine::default()`


mohanson:
Since we have two different versions now, and the supported instruction sets include IMC, B, MOP, `latest()` and `default()` bring more confusion than their benefits, so I recommend removing them.

Related code:

- https://github.com/nervosnetwork/ckb-vm/blob/develop/src/machine/mod.rs#L365-L371
- https://github.com/nervosnetwork/ckb-vm/blob/develop/tests/test_asm.rs#L26

xxuejie:
I think we can pull the trigger here. I don't have any arguments for having them in the API now

mohanson:
Merged.

PR #165, opened 2021-06-02
feat: compile aot code in AsmMachine.load_elf

mohanson:
Rewrite the logic of aot machine when processing reset, now the second program can also be compiled to aot code normally.

- Add a new structure AsmWrapMachine
- `instruction_cycle_func` is now used as a parameter of `AotCompilingMachine.compile()` instead of a parameter of `AotCompilingMachine::new()`
- Remove confusing constructor like machine::latest() or machine::default(): https://github.com/nervosnetwork/ckb-vm/issues/163

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,7 +8,7 @@ trigger:
 
 variables:
   RUST_BACKTRACE: full
-  TEST_SUITE_COMMIT: dea02067247a17e5d5059485e3f8ccb5c92a6810
+  TEST_SUITE_COMMIT: 8c0c2239194dc283308377c3fd4e04744b01c988
 
 jobs:
   - job: WinUnitTest
--- a/benches/vm_benchmark.rs
+++ b/benches/vm_benchmark.rs
@@ -1,52 +1,64 @@
 #[macro_use]
 extern crate criterion;
 
-use bytes::Bytes;
-#[cfg(has_asm)]
-use ckb_vm::machine::{
-    aot::AotCompilingMachine,
-    asm::{AsmCoreMachine, AsmMachine},
-    DefaultMachineBuilder, VERSION0, VERSION1,
+use ckb_vm::{
+    machine::VERSION0, memory::wxorx::WXorXMemory, Bytes, DefaultCoreMachine,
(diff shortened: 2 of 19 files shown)


xxuejie:
Can we split this into 2 PRs? The changes for `latest` and `default` could go first, but the change about `AsmWrapMachine` is so weird, I do think we should spend more time on that.

mohanson:
Okay, I will split this into two PRs, and then explain why I want to add AsmWrapMachine

mohanson:
image

mohanson:
I think the core problem is that we can't compile aot code in syscall, just like we can't clean Trace in syscall.

xxuejie:
> we can't compile aot code in syscall

This is an interesting one: why can't we compile aot code within a syscall? What if we just create a new AOT machine, and use it to compile aot code?

mohanson:
>  What if we just create a new AOT machine, and use it to compile aot code?

First question: How do we know that we are running in aot mode in syscall? 

And `instruction_cycle_func` is needed to compile aot code, this function is not available in syscall.

The last: How to pass the compiled result to the outside? Only `Box<AsmCoreMachine>` can be accessed in syscall as a `SupportMachine`, and it does not have the ability to save aot_code

xxuejie:
> > What if we just create a new AOT machine, and use it to compile aot code?
> 
> First question: How do we know that we are running in aot mode in syscall?
> 

I consider syscalls together with the VM initialization code as a unified part. After all, syscall impls are put into VM when initializing the VM. When you are initializing VM, you know for a fact if you use AOT mode, in this sense, a syscall module should already have the information whether current VM uses AOT mode or not.

> And `instruction_cycle_func` is needed to compile aot code, this function is not available in syscall.

Both this function, and the actual syscall modules, are provided by the initialization code to the VM, so it's natural that we can feed this function to a syscall module as well.

> 
> The last: How to pass the compiled result to the outside? Only `Box<AsmCoreMachine>` can be accessed in syscall as a `SupportMachine`, and it does not have the ability to save aot_code

This will require some careful thinking, I think we need an interface for putting auxillary data or sorts to a SupportMachine instance.

PR #166, opened 2021-06-09
refactor: remove constructor default() and latest()

mohanson:
ref #163

### DIFF
--- a/definitions/src/generate_asm_constants.rs
+++ b/definitions/src/generate_asm_constants.rs
@@ -110,7 +110,7 @@ fn main() {
         size_of::<AsmCoreMachine>()
     );
 
-    let m: Box<AsmCoreMachine> = Box::<AsmCoreMachine>::default();
+    let m: Box<AsmCoreMachine> = AsmCoreMachine::new(0, 0, 0);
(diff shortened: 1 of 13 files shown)


xxuejie (dismissed):

definitions/src/generate_asm_constants.rs
```diff
@@ -110,7 +110,7 @@ fn main() {
         size_of::<AsmCoreMachine>()
     );
 
-    let m: Box<AsmCoreMachine> = Box::<AsmCoreMachine>::default();
+    let m: Box<AsmCoreMachine> = AsmCoreMachine::new(0, 0, 0);
```
doitian, line 113: From this invocation, it is very easy to pass parameters in the wrong positions. Maybe it requires a builder as well.

doitian (dismissed):

definitions/src/generate_asm_constants.rs
mohanson: There are not many parameters at present, if there are more than 5 parameters, then it is reasonable to add a builder like `AsmCoreMachineBuilder`

xxuejie (approved):

Issue #169, opened 2021-06-14
Macro-op-fusion: Pattern design of ADC and SBB


mohanson:
# ADC

Adding a0, a1, and a2 with results in a0 and carry-out in a1.

```
add  a0, a0, a1
sltu a1, a0, a1
add  a0, a0, a2
sltu a2, a0, a2
or   a1, a1, a2
```

- a0 ~ a2 are used as placeholders
- a0, a1, a2 != ZERO

# SBB

Do a0 - a1 - a2 with results in a0 and carry-out in a1.

```
sub  a1, a0, a1
sltu a3, a0, a1
sub  a0, a1, a2
sltu a2, a1, a0
or   a1, a2, a3
```

- a0 ~ a3 are used as placeholders
- a3 is a temporary register
- a0, a1, a2, a3 != ZERO

We need to read 5 instructions to really determine whether these 5 instructions can be fused into a mop instruction, but in fact, when the first instruction is read, we have a high accuracy rate to judge whether the first instruction is the beginning of an sbb mop instructions. This way `sub a1, a0, a1` of writing `sub` is not very common, in most cases we will write like this `sub a0, a0, a1`. Therefore, this effectively distinguishes `sbb` and normal `sub`.

Test code: https://github.com/mohanson/ckb-vm/tree/adcsbb

mohanson:
cc @XuJiandong @xxuejie

mohanson:
Features have been merged by #170

PR #170, opened 2021-06-14
feat: macro-op fusion adc / sbb

mohanson:
ref: #169


src/decoder.rs
```diff
@@ -101,14 +101,195 @@ impl Decoder {
...
+                    Ok(i)
+                } else {
+                    Ok(head_instruction)
+                }
+            }
+            insts::OP_SUB => {
+                let mut rule_sbb = || -> Result<Option<Instruction>, Error> {
+                    let head_inst = Rtype(head_instruction);
+                    let head_size = instruction_length(head_instruction);
+                    if head_inst.rs1() == ZERO || head_inst.rs2() == ZERO {
```
xxuejie: For adc, we were checking rd of `head_inst` is not ZERO, why this is missing here?

xxuejie: I think it might be better to have a unified check, where we are checking all 3 registers used are not ZERO.

src/machine/asm/execute.S
```diff
@@ -2191,6 +2197,37 @@ ckb_vm_x64_execute:
   WRITE_RS3(RS2r)
   NEXT_INST
 .p2align 3
+.CKB_VM_ASM_LABEL_OP_ADC:
```
xxuejie, line 2200: This is a direct translation of the 5 instructions from RISC-V, but from a semantic point of view, what we need to do here, is simply:

* Set or clear carry flag based on passed value from RISC-V register(there are many [ways](https://stackoverflow.com/questions/36798572/how-can-i-set-or-clear-overflow-flag-in-x86-assembly) to do it)
* Use adc on the 2 source RISC-V registers
* Move carry flag to the output register

This flow, IMO, really matches what ADC is really doing.

tests/test_mop.rs
```diff
@@ -143,3 +143,47 @@ pub fn test_mop_secp256k1() {
...
+    #[cfg(has_asm)]
+    {
+        let mut machine_asm = machine_build::asm_v1_mop("tests/programs/mop_adc", vec![]);
+        let ret_asm = machine_asm.run();
+        assert!(ret_asm.is_ok());
+        assert_eq!(ret_asm.unwrap(), 0);
+
+        let code = machine_build::aot_v1_mop_code("tests/programs/mop_adc");
+        let mut machine_aot = machine_build::aot_v1_mop("tests/programs/mop_adc", vec![], &code);
+        let ret_aot = machine_aot.run();
```
xxuejie, line 171: It's best if we add test cases validating cycles consumed by the mop version vs non-mop version.

src/decoder.rs
mohanson: The next codes will check head_inst.rd() == head_inst.rs2()

xxuejie: Yep but we are relying an inherit rule that is quite prone to errors, I do suggest the following changes:

* When we were decoding instructions, we only check for the equality of certain register IDs, such as `head_inst.rd() == head_inst.rs2()`
* When we have decoded all the instructions, and are ready to assemble the mop adc/sbb instruction, we add a ZERO register check for all used registers by the mop instruction

I think this way we can have better code readability, and I do suspect the generated optimized code, will be the same level of performance.

src/machine/asm/execute.S
mohanson: I can't get the correct a2 value under the steps you recommend. Consider the result of a2, I think direct translation is the most convenient. 

```
a2 = (tmp + a2) 's carry flag
where tmp = a0 + a1
```

src/decoder.rs
mohanson: Yes, this way is better

tests/test_mop.rs
mohanson: done

xxuejie (approved):

PR #175, opened 2021-07-05
Add method add_cycles_no_checking

mohanson:


xxuejie (dismissed):

tests/test_aot.rs
```diff
@@ -375,3 +375,48 @@ pub fn test_aot_rvc_pageend() {
...
+        .syscall(Box::new(OutOfCyclesSyscall {}))
+        .build();
+    let mut machine = AsmMachine::new(core, Some(&code));
+    machine
+        .load_program(&buffer, &vec!["syscall".into()])
+        .unwrap();
+    let result = machine.run();
+    assert!(result.is_err());
+    assert_eq!(result.unwrap_err(), Error::InvalidCycles);
+    assert_eq!(machine.machine.cycles(), 108);
```
xxuejie, line 421: We should add an assertion line here, ensuring `A0` is indeed overwritten, meaning the syscall completes execution despite the cycle error

xxuejie (approved):

PR #179, opened 2021-07-20
perf: optimize page_indices bound check

quake:
run 5 times 
```
0.19.4      2.2706 ms
0.20.0-rc1  2.5129 ms
this pr     2.4360 ms
```

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -91,12 +91,20 @@ pub extern "C" fn inited_memory(frame_index: u64, machine: &mut AsmCoreMachine) 
 }
 
 fn check_memory(machine: &mut AsmCoreMachine, page_indices: &(u64, u64)) {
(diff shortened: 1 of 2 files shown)


mohanson:
For the instruction execution, the memory can only be read and written at most 8 bytes at a time, so can it be assumed that the fact is true: During the execution of the instruction, if it read/write memory from more than one frame, then the frame_end = frame + 1 . In this way, all `iter()` operations can be avoided.

quake:
> For the instruction execution, the memory can only be read and written at most 8 bytes at a time, so can it be assumed that the fact is true: During the execution of the instruction, if it read/write memory from more than one frame, then the frame_end = frame + 1 . In this way, all `iter()` operations can be avoided.

resolved in https://github.com/nervosnetwork/ckb-vm/pull/183

PR #180, opened 2021-07-21
perf: add decoder instructions cache

quake:
After reading this [comment](https://github.com/nervosnetwork/ckb-vm/pull/179#issuecomment-883460530), I realized that the real performance bottleneck is in the `execute_load32` and `execute_load16` method, which could be improved by adding cache. 

Since we have implemented W^X memory already, I think a simple FIFO cache is enough for the decoder, the results of the benchmark tests look good:

run 5 times
```
0.19.4      2.2706 ms
0.20.0-rc1  2.5129 ms
this pr     2.2421 ms
```

and fyi, I tested different cache size from `16` to `8192`, 4096 is best on my local machine.


src/decoder.rs
```diff
@@ -80,10 +83,16 @@ impl Decoder {
         }
     }
 
-    pub fn decode_raw<M: Memory>(&self, memory: &mut M, pc: u64) -> Result<Instruction, Error> {
+    pub fn decode_raw<M: Memory>(&mut self, memory: &mut M, pc: u64) -> Result<Instruction, Error> {
+        let instruction_cache_key = pc as usize % INSTRUCTION_CACHE_SIZE;
```
xxuejie: Due to RISC-V instruction encoding, the lowest bit in PC will always be zero, so one trivial work we can do is:

`(pc >> 1) as usize % INSTRUCTION_CACHE_SIZE`

Otherwise half of the cache will always be empty.

xxuejie (approved):

PR #183, opened 2021-07-27
perf: tweak memory related check of AsmMachine

quake:

5 times avg:
```
develop: 2.2431 ms
this pr: 2.2265 ms
```

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -3,14 +3,15 @@ use std::mem::transmute;
 use byteorder::{ByteOrder, LittleEndian};
 use bytes::Bytes;
 pub use ckb_vm_definitions::asm::AsmCoreMachine;
-use ckb_vm_definitions::asm::{
-    calculate_slot, Trace, RET_CYCLES_OVERFLOW, RET_DECODE_TRACE, RET_DYNAMIC_JUMP, RET_EBREAK,
-    RET_ECALL, RET_INVALID_PERMISSION, RET_MAX_CYCLES_EXCEEDED, RET_OUT_OF_BOUND, RET_SLOWPATH,
-    TRACE_ITEM_LENGTH, TRACE_SIZE,
-};
 use ckb_vm_definitions::{
-    instructions::OP_CUSTOM_TRACE_END, ISA_MOP, MEMORY_FRAMES, MEMORY_FRAME_PAGE_SHIFTS,
-    RISCV_GENERAL_REGISTER_NUMBER,
+    asm::{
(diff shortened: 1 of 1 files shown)


src/machine/asm/mod.rs
```diff
@@ -101,6 +102,97 @@ fn check_memory(machine: &mut AsmCoreMachine, page_indices: &(u64, u64)) {
...
+    addr: u64,
+    size: usize,
+) -> Result<(), Error> {
+    debug_assert!(size == 1 || size == 2 || size == 4 || size == 8);
+    let page = addr >> RISCV_PAGE_SHIFTS;
+    if page as usize >= RISCV_PAGESIZE {
+        return Err(Error::OutOfBound);
+    }
+
+    let page_offset = addr as usize % RISCV_PAGESIZE;
```
mohanson, line 126: May be faster if the compiler does not do this optimization : `let page_offset = addr as usize & 0xfff`

src/machine/asm/mod.rs
```diff
@@ -101,6 +102,97 @@ fn check_memory(machine: &mut AsmCoreMachine, page_indices: &(u64, u64)) {
...
+        }
+
+        let frame = page >> MEMORY_FRAME_PAGE_SHIFTS;
+        if machine.frames[frame as usize] == 0 {
+            inited_memory(frame, machine);
+            machine.frames[frame as usize] = 1;
+        }
+
+        machine.set_flag(page, FLAG_DIRTY)?
+    } else {
```
mohanson: If a `write` method use two pages, then two pages should be checked, but you only check the latter page here.

src/machine/asm/mod.rs
```diff
@@ -198,15 +290,17 @@ impl Memory for Box<AsmCoreMachine> {
...
-        let page_indices = get_page_indices(addr, 4)?;
-        check_permission(self, &page_indices, FLAG_EXECUTABLE)?;
-        self.load32(&(addr)).map(|v| v as u32)
+        check_memory_executable(self, addr, 4)?;
+        Ok(LittleEndian::read_u32(
+            &self.memory[addr as usize..addr as usize + 4],
+        ))
     }
 
     fn load8(&mut self, addr: &u64) -> Result<u64, Error> {
```
mohanson, line 309: It would be great if you can also modify load8/16/32/64

quake: resolved, please review again.

src/machine/asm/mod.rs
quake: sorry for the mistake, please review again.

src/machine/asm/mod.rs
quake: I can confirm that the compiler will optimize the code, will keep it unchanged

mohanson (approved):
LGTM

PR #201, opened 2021-08-20
Fix: out of bounds check for pages

mohanson:


quake (approved):

xxuejie (approved):
Should we create a new release immediately with the fix here?

mohanson:
> Should we create a new release immediately with the fix here?

I am still fuzzing exec syscall / ckb-vm, next Monday is a good time

PR #203, opened 2021-09-01
feat: assembly mode implementation for aarch64 architecture

xxuejie:
Based on initial tests on AWS hardware, this PR could boost CKB on aarch64 architecture to similar performance on x64 architecture.

Note that at the moment, CKB does not use CKB-VM's AOT mode, meaning when this PR is done, we can have a performant CKB implementation on aarch64, at least from a VM's perspective.


xxuejie:
This is now ready for review, we have added asm mode for aarch64 architecture. The CI is also updated so we are running unit tests & test suite in a qemu environment(Azure Pipeline does not have native arm64 machines for us to test)

src/machine/asm/execute_aarch64.S
```diff
@@ -0,0 +1,1634 @@
...
+#define RS2 x18
+#define RS2w w18
+#define RS3 x19
+#define RS3w w19
+#define IMMEDIATE x20
+#define IMMEDIATEw w20
+#define FLAG x21
+
+#define REGISTER_BASE x22
+#define ZERO_VALUE x23
```
mohanson: It's better to use `xzr` register and then remove the `mov ZERO_VALUE 0`

src/machine/asm/execute_aarch64.S
```diff
@@ -0,0 +1,1633 @@
...
+  mov TEMP4, 0
+  mov TEMP3, 0
+.clmul_branch:
+  mov TEMP1, RS1
+  mov TEMP2, RS2
+  lsl TEMP1, TEMP1, TEMP4
+  lsr TEMP2, TEMP2, TEMP4
+  eor TEMP1, TEMP1, TEMP3
+  and TEMP2, TEMP2, 1
+  csel TEMP3, TEMP1, TEMP3, ne
```
mohanson, line 1074: It's necessary to you to add `cmp TEMP2, 0` before `csel`.

Same bugs in `clmulh` and `clmulr`

Test case

```
.global _start
_start:
    li a0, 10
    li a1, 20
    clmul a0, a0, a1
    add a0, a0, 120
    li a7, 93
    ecall
```

src/machine/asm/execute_aarch64.S
```diff
@@ -0,0 +1,1636 @@
...
+  WRITE_RD(RS1)
+  NEXT_INST
+.CKB_VM_ASM_LABEL_OP_ORCB:
+  DECODE_R
+  ldr RS1, REGISTER_ADDRESS(RS1)
+  mov RS2, 0
+  ldr TEMP1, =0x00000000000000ff
+  mov TEMP2, TEMP1
+  and TEMP1, TEMP1, RS1
+  beq .orcb_branch1
```
mohanson, line 1248: Add `cmp TEMP1, 0` before `beq .orcb_branch1`, `beq .orcb_branch2` ... `beq .orcb_branch8`

TestCase:

```
.global _start
_start:
    li a1, 0xffffffffffffee00
    orc.b a0, a1
    li a2, 0xffffffffffffff00
    beq a0, a2, ok
    li a0, 1
    li a7, 93
    ecall
ok:
    li a0, 0
    li a7, 93
    ecall
```

src/machine/asm/execute_aarch64.S
```diff
@@ -0,0 +1,1644 @@
...
+  str RS1, PC_ADDRESS
+  b .prepare_trace
+.CKB_VM_ASM_LABEL_OP_DIV:
+  DECODE_R
+  mov TEMP1, INT64_MIN
+  ldr RS1, REGISTER_ADDRESS(RS1)
+  ldr RS2, REGISTER_ADDRESS(RS2)
+  cmp RS1, TEMP1
+  bne .div_branch1
+  cmp RS2, -1
```
mohanson: Not a bug, but we may be able to cut the `INT_MIN/-1`  branch: 

https://developer.arm.com/documentation/den0024/a/The-A64-instruction-set/Data-processing-instructions/Multiply-and-divide-instructions

```
Overflow can only occur in SDIV:
INT_MIN / -1 returns INT_MIN, where INT_MIN is the smallest negative number that can be encoded in the registers used for the operation. The result is always rounded towards zero, as in most C/C++ dialects.
```

mohanson (approved):
LGTM

PR #210, opened 2021-09-27
fix: fuzz build failed in latest nightly rust

mohanson:


driftluo:
https://github.com/rust-fuzz/cargo-fuzz/pull/277 When the next version is released, this pr needs to be reverted 

it has been published

PR #228, opened 2021-12-21
Add vand.*, vor.*, vxor.*

XuJiandong:


XuJiandong:
See test cases here:
https://github.com/mohanson/rvv-playground/pull/4/files

PR #234, opened 2021-12-30
Refactor Error

mohanson:
- Optimize error's fmt output, making it easier for users to debug.

- Distinguish between 
    - `AotLimitReachedMaximumTempRegisters`, 
    - `AotLimitReachedMaximumSections`, 
    - `AotLimitReachedMaximumLabels`, 
    - `AotLimitReachedMaximumDummySections`, 
    - `AotSectionIsEmpty`, 
    - `AotSectionOverlaps`, 
    - `AotOutOfBoundDueToNotStartOfBasicBlock`, 
    - `ElfSegmentAddrOrSizeError`, 
    - `MemOutOfBound`, 
    - `MemOutOfStack` 
    from `OutOfBound`

- Distinguish between 
    - `MemWriteOnExecutablePage`, 
    - `MemWriteOnFreezedPage`, 
    - `ElfSegmentUnreadable`, 
    - `ElfSegmentWritableAndExecutable`
    from `InvalidPermission`

- Rename `InvalidCycles` to `CyclesExceeded`
- Rename `Unaligned` to `MemPageUnalignedAccess`
- Rename `Dynasm` to `AotDynasm`
- Rename `InvalidElfBits` to `ElfBits`

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -64,7 +64,7 @@ pub enum Error {
     #[display(fmt = "memory error: write on freezed page")]
     MemWriteOnFreezedPage,
     #[display(fmt = "unexpected error")]
-    Unexpected,
+    Unexpected(String),
     #[display(fmt = "unimplemented")]
     Unimplemented,
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -12,6 +12,10 @@ pub enum Error {
     Asm(u8),
     #[display(fmt = "elf error: {}", "_0")]
     ElfParseError(String),
+    #[display(fmt = "elf error: segments is unreadable")]
+    ElfSegmentUnreadable,
+    #[display(fmt = "elf error: segments is writable and executable")]
+    ElfSegmentWritableAndExecutable,
     #[display(fmt = "I/O error: {:?}", "_0")]
     IO(std::io::ErrorKind),
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -22,6 +22,10 @@ pub enum Error {
     ElfSegmentUnreadable,
     #[display(fmt = "elf error: segments is writable and executable")]
     ElfSegmentWritableAndExecutable,
+    // Unknown error type is for the debugging tool of CKB-VM, it should not be
+    // used in this project.
+    #[display(fmt = "external error: {}", "_0")]
+    External(String),
     #[display(fmt = "invalid syscall {}", "_0")]
     InvalidEcall(u64),
     #[display(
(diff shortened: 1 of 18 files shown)


src/error.rs
```diff
@@ -1,56 +1,87 @@
...
+    #[display(fmt = "aot error: limit reached maximum labels")]
+    AotLimitReachedMaximumLabels,
+    #[display(fmt = "aot error: limit reached maximum sections")]
+    AotLimitReachedMaximumSections,
+    #[display(fmt = "aot error: limit reached maximum temp register")]
+    AotLimitReachedMaximumTempRegisters,
+    #[display(fmt = "aot error: out of bound due to not start of basic block")]
+    AotOutOfBoundDueToNotStartOfBasicBlock,
+    #[display(fmt = "asm error: {}", "_0")]
+    Asm(u8),
```
xxuejie, line 20: Where is this error used?

src/error.rs
```diff
@@ -1,56 +1,87 @@
...
+    ElfBits,
+    #[display(fmt = "elf error: {}", "_0")]
+    ElfParseError(String),
+    #[display(fmt = "elf error: segment is unreadable")]
+    ElfSegmentUnreadable,
+    #[display(fmt = "elf error: segment is writable and executable")]
+    ElfSegmentWritableAndExecutable,
+    #[display(fmt = "elf error: segment addr or size is wrong")]
+    ElfSegmentAddrOrSizeError,
+    // Unknown error type is for the debugging tool of CKB-VM, it should not be
```
xxuejie: This comment says "Unknown", but the error below is named "External"

src/error.rs
```diff
@@ -1,56 +1,87 @@
...
-    #[display(fmt = "assembly error {}", "_0")]
-    Asm(u8),
-    #[display(fmt = "limit reached")] // FIXME: Distinguish which limit
-    LimitReached,
-    #[display(fmt = "invalid permission")] // FIXME: Distinguish which permission
-    InvalidPermission,
     #[display(fmt = "invalid version")]
     InvalidVersion,
+    #[display(fmt = "I/O error: {:?}", "_0")]
+    IO(std::io::ErrorKind),
```
xxuejie: Will it be better if we preserve both the error kind, and a raw error message as string?

src/error.rs
mohanson: https://github.com/mohanson/ckb-vm/blob/error/src/machine/asm/mod.rs#L537

The `Asm` error will not be triggered unless there are bugs in the asm code

src/error.rs
```diff
@@ -1,56 +1,87 @@
...
+    #[display(fmt = "memory error: out of stack")]
+    MemOutOfStack,
+    #[display(fmt = "memory error: unaligned page access")]
+    MemPageUnalignedAccess,
+    #[display(fmt = "memory error: write on executable page")]
+    MemWriteOnExecutablePage,
+    #[display(fmt = "memory error: write on freezed page")]
+    MemWriteOnFreezedPage,
     #[display(fmt = "unexpected error")]
     Unexpected,
```
XuJiandong: need more information about this error. what about  Unexpected(String)?

XuJiandong (changes requested):

src/error.rs
mohanson: I will modify it.

src/error.rs
mohanson: Agree.

src/error.rs
mohanson: Good idea.

xxuejie (approved):

chaoticlonghair:
LGTM, except the follow proposal:

- Since `RET_INVALID_PERMISSION -> Error::MemWriteOnExecutablePage` is a injection:
  https://github.com/nervosnetwork/ckb-vm/blob/66799c6a4781a8d96677a660373f8fa04edf5453/src/machine/asm/mod.rs#L531
  https://github.com/nervosnetwork/ckb-vm/blob/66799c6a4781a8d96677a660373f8fa04edf5453/src/machine/asm/mod.rs#L578
  Maybe `*_INVALID_PERMISSION` should be updated to `*_MEM_WRITE_ON_EXECUTABLE_PAGE`, such as:
  https://github.com/nervosnetwork/ckb-vm/blob/66799c6a4781a8d96677a660373f8fa04edf5453/definitions/src/asm.rs#L18
  https://github.com/nervosnetwork/ckb-vm/blob/66799c6a4781a8d96677a660373f8fa04edf5453/src/machine/asm/cdefinitions_generated.h#L20

mohanson:
Yes, this naming can be optimized

PR #243, opened 2022-02-28
chore: print on build command error

doitian:
Print stdout and stderr on build command error.

By default, `std::process::Command` only prints stderr on error. Stdout may help to figure out why `yasm` failed in CI like in https://github.com/nervosnetwork/ckb/runs/5343852953?check_suite_focus=true


mohanson:
LGTM

mohanson (approved):

PR #250, opened 2022-03-14
RVV: fix bug vslidexxx and vcompress.vm

joii2020:
vslidexxx: can not get mask
vcompress.vm: vs legth get error


doitian:
If you think there's a bug, please give some test cases or examples to show the bug.

joii2020:
> If you think there's a bug, please give some test cases or examples to show the bug.


https://github.com/joii2020/rvv-testcases/blob/dev/cases/src/vector_slide_cases.rs
https://github.com/joii2020/rvv-testcases/blob/dev/cases/src/vector_compress_cases.rs

mohanson (approved):

PR #253, opened 2022-03-22
Fix vrgather.vv, when sew more than 64

joii2020:
When sew more than 64, high-order data may be lost.


joii2020:
This code can be trigger:
https://github.com/joii2020/rvv-testcases/blob/master/cases/src/vector_register_gather_cases.rs#L217

PR #263, opened 2022-06-01
Save program bytes in DefaultCoreMachine

mohanson:
The caller now has a chance to get the code loaded via the exec syscall.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ aot = ["asm"]
 enable-chaos-mode-by-default = ["ckb-vm-definitions/enable-chaos-mode-by-default"]
 # Disable slow tests to run miri on CI
 miri-ci = []
+pprof = []
 
 [dependencies]
 byteorder = "1"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -209,7 +209,9 @@ jobs:
       - script: |
           sudo apt-get install -y qemu binfmt-support qemu-user-static
           docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
-          docker run --rm -v `pwd`:/code -t arm64v8/rust bash -c "cd /code && cargo test --features=asm"
+          echo "Comment: updating crates.io cause oom in arm docker, use -v /usr/local/cargo/registry/:/usr/local/cargo/registry/ to skip this problem"
+          cargo update
+          docker run --rm -v `pwd`:/code -v /home/vsts/.cargo/registry/:/usr/local/cargo/registry/ -t arm64v8/rust bash -c "cd /code && cargo test --features=asm"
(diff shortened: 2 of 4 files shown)


mohanson:
cc @xxuejie @XuJiandong

XuJiandong (approved):

xxuejie:
Judging from the code I cannot see what the proper use case is here, is it possible to provide some more background here?

mohanson:
> Judging from the code I cannot see what the proper use case is here, is it possible to provide some more background here?

ckb-debugger does no handle exec syscall properly, due to missing raw elf bytes. After this PR, we can solve the problem by:

https://github.com/mohanson/ckb-vm-pprof/commit/14673052632d5ecbbe1e6a22471dccf2fb8ebe2b

xxuejie (approved):

quake:
> > Judging from the code I cannot see what the proper use case is here, is it possible to provide some more background here?
> 
> ckb-debugger does no handle exec syscall properly, due to missing raw elf bytes. After this PR, we can solve the problem by:
> 
> [mohanson/ckb-vm-pprof@1467305](https://github.com/mohanson/ckb-vm-pprof/commit/14673052632d5ecbbe1e6a22471dccf2fb8ebe2b)

I don't quite understand the purpose of this PR, if it's just for profiler to load data, wouldn't it be enough to add a field to PProfMachine?

```rust
pub struct PProfMachine {
    pub machine: DefaultMachine,
    pub profile: Profile,
    pub program: Bytes,
}

impl PProfMachine {
  pub fn load_program() {
      self.program = program.clone();
      self.machine.load_program(program, args)
  }
}
```

mohanson:
> > > Judging from the code I cannot see what the proper use case is here, is it possible to provide some more background here?
> > 
> > 
> > ckb-debugger does no handle exec syscall properly, due to missing raw elf bytes. After this PR, we can solve the problem by:
> > [mohanson/ckb-vm-pprof@1467305](https://github.com/mohanson/ckb-vm-pprof/commit/14673052632d5ecbbe1e6a22471dccf2fb8ebe2b)
> 
> I don't quite understand the purpose of this PR, if it's just for profiler to load data, wouldn't it be enough to add a field to PProfMachine?
> 
> ```rust
> pub struct PProfMachine {
>     pub machine: DefaultMachine,
>     pub profile: Profile,
>     pub program: Bytes,
> }
> 
> impl PProfMachine {
>   pub fn load_program() {
>       self.program = program.clone();
>       self.machine.load_program(program, args)
>   }
> }
> ```

It's about profiler loading program data by **exec syscall**, there is no way of knowing what was loaded by exec syscall unless the syscall was executed.

quake:
Got, thanks.

PR #271, opened 2022-07-18
Calculate trace slot by shift 2

mohanson:


mohanson:
When run `./ckb replay --tmp-target tmp --profile 7000000 7100000` in mainnet,

- ckb v0.103.0 took 47 hours 30 minutes
- this patch only took 25 hours and 15 minutes

XuJiandong (approved):

PR #274, opened 2022-08-06
perf: Prefetch consecutively the next trace to L3 cache

xxuejie:
It is not always correct in the presence of branches, but when ckb-vm is running sequential code, this could help alleviate memory reads by a large marge.

This brings about 5% performance improvements per my tries.


mohanson (approved):
LGTM

PR #276, opened 2022-08-09
chore: use Arc instead of lifetime parameter

mohanson:
This PR modified the definition of AsmMachine:

```
// Before
pub struct AsmMachine<'a> {
    pub machine: DefaultMachine<'a, Box<AsmCoreMachine>>,
    pub aot_code: Option<&'a AotCode>,
}

// After
pub struct AsmMachine {
    pub machine: DefaultMachine<Box<AsmCoreMachine>>,
    pub aot_code: Option<Arc<AotCode>>,
}
```

This allows us to completely remove any lifetime parameter in ckb-vm.

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -1,5 +1,17 @@
-use std::mem::transmute;
-
+use crate::{
+    decoder::{build_decoder, Decoder},
+    instructions::{
+        blank_instruction, execute_instruction, extract_opcode, instruction_length,
+        is_basic_block_end_instruction,
+    },
+    machine::VERSION0,
+    memory::{
+        fill_page_data, get_page_indices, memset, round_page_down, round_page_up, FLAG_DIRTY,
+        FLAG_EXECUTABLE, FLAG_FREEZED, FLAG_WRITABLE, FLAG_WXORX_BIT,
+    },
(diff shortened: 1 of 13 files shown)


src/machine/asm/mod.rs
```diff
@@ -433,9 +432,9 @@ impl AotCode {
     }
 }
 
-pub struct AsmMachine<'a> {
-    pub machine: DefaultMachine<'a, Box<AsmCoreMachine>>,
-    pub aot_code: Option<&'a AotCode>,
+pub struct AsmMachine {
+    pub machine: DefaultMachine<Box<AsmCoreMachine>>,
+    pub aot_code: Option<Rc<AotCode>>,
```
xxuejie: I think it's better to use `Arc` here. It's not that ckb-vm will be using threads, it's more that if you think about the usage patterns, it's likely the same binary being AOTed could be shared amongst multiple ckb-vm instances, all using the same `AotCode` instance to speed up execution. In this case, `AotCode` might be shared among multiple threads. Using `Rc` here severely limits the ability to share compiled immutable code.

mohanson: Right.

mohanson:
CC @xxuejie

xxuejie (approved):

PR #277, opened 2022-08-10
feat: Support detect-aot feature

jjyr:
To support compiling on Mac m1, we want to enable the `aot` feature depend on the compiling environment. Unforturnately, rust doesn't allow us to turn on a feature in `build.rs`.

This PR adds `detect-aot` feature, just like the `detect-asm` - the AOT is enabled when it is possible.


mohanson (approved):
LGTM

PR #282, opened 2022-08-15
Fix bug in vsaddu.vi

joii2020:
Fix the bug that vsaddu.vi gets imm.
The [testcase](https://github.com/XuJiandong/rvv-testcases/blob/master/cases/src/single_saturating_add_subtract_cases.rs#L243)
Update eint


mohanson:
But the diff shows that you modified vsadd**u**_vi

joii2020:
> 

Sorry, typo, changed

PR #288, opened 2022-09-07
feat: Replace LD_SIGN_EXTENDED_32_CONSTANT with CUSTOM_LOAD_UIMM

xxuejie:
LD_SIGN_EXTENDED_32_CONSTANT is exactly the same as CUSTOM_LOAD_IMM, this change removes LD_SIGN_EXTENDED_32_CONSTANT, then adds CUSTOM_LOAD_UIMM, which as shown in the test case , can be helpful to eliminate all AUIPCs.


mohanson (approved):
LGTM

PR #295, opened 2022-10-10
refactor: syscall mut ref

quake:


quake:
close this pr since it will break the lifetime requirement, for example:
https://github.com/nervosnetwork/ckb/blob/7672bc7164f1ff82d5d143609c9d0f8e3543c2fb/script/src/types.rs#L101

PR #296, opened 2022-10-20
Remove aot codes

mohanson:
In this PR, we moved the AOT implementation of ckb-vm to a new project:

https://github.com/mohanson/ckb-vm-aot


xxuejie:
Before jumping into actual review: the README file contains a section describing different running modes which also mentions the AOT mode, please revise the README file as well.

mohanson:
> Before jumping into actual review: the README file contains a section describing different running modes which also mentions the AOT mode, please revise the README file as well.

Fixed

src/instructions/ast.rs
```diff
@@ -1,445 +0,0 @@
-use crate::Register;
```
xxuejie: I do recommend we keep this file around. It could be useful in numerous tools surrounding ckb-vm, while also providing an example that the register value of a VM instance could be customized.

xxuejie:
Without a way to generate `AotCode` in this repo, I believe we should also remove this: https://github.com/nervosnetwork/ckb-vm/blob/0ca74df911b7a67aad72c88a7a35109505c85515/src/machine/asm/mod.rs#L438 It could be confusing when AsmMachine accepts `AotCode` struct but does not provide a reasonable way to generate it.

Cargo.toml
```diff
@@ -37,13 +34,13 @@ rand = "0.7.3"
 # Feature detection won't work here
 [target.'cfg(any(windows, unix))'.dependencies]
 libc = "0.2.47"
-memmap = { package ="mapr", version = "0.8.0" }
+memmap = { package ="memmap2", version = "0.5" }
```
xxuejie: With aot removed, I don't think we would be needing those 2 crates anymore, correct?

src/instructions/ast.rs
mohanson: Done

Cargo.toml
mohanson: That's right

xxuejie: I was actually referring to **2** crates here, we can also safely remove libc. There's one extra reference of `c_uchar` in the code but std [has](https://doc.rust-lang.org/std/os/raw/type.c_uchar.html) it already.

mohanson: Done, libc has also been removed

xxuejie (approved):

mohanson:
CI seems to be stuck, can you force merge this PR? @xxuejie

xxuejie:
@mohanson It's actually github's branch protection that is requiring certain jobs that are now cleaned. I've revised branch protection and now merging should work.

PR #300, opened 2022-10-26
Utlize instruction cache for decode_mop

XiaowenHu96:
The instruction cache was not enabled for decode_mop, this pr fixes it.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -117,9 +117,18 @@ impl Decoder {
     // - https://en.wikichip.org/wiki/macro-operation_fusion#Proposed_fusion_operations
     // - https://carrv.github.io/2017/papers/clark-rv8-carrv2017.pdf
     pub fn decode_mop<M: Memory>(&mut self, memory: &mut M, pc: u64) -> Result<Instruction, Error> {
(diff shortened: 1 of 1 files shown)


mohanson:
I doubt whether the `instructions_cache` is necessary. Considering that the trace hit is already as high as 98%, I think it is possible to completely remove the instructions_cache in the decoder.

XiaowenHu96:
@mohanson 
It is only necessary because the interpreter mode does not enable TraceMachine. Another way of fixing it would be just to let the interpreter use the trace machine.
Otherwise, the current decode method is efficient enough. But as requested, I'll have a look into a more efficient decoding method.

BTW the test is failing, I'll have a look.

xxuejie:
I think it's best to consult on actual numbers. Back in the days when #180 is introduced, instruction cache does help with performance. Decoding still takes time even when we have a higher trace hit ratio(you still need to decode each instruction when running it for the first time)

XiaowenHu96:
The test fails because mop and raw have to use different cache lines. Otherwise, calls to `decode_raw` within `decode_mop` will corrupt the cache result. I'll close this PR for now.

Issue #302, opened 2022-11-02
Discussion: Remove duplicate out of bounds checks


mohanson:
In the current assembly code, every Load instruction needs to go through 3(or 4) out_of_bound checks:

```text
#define CHECK_READ_BOUND_VERSION1(length) \
  movq REGISTER_ADDRESS(RS1), RS1; \
  addq IMMEDIATE, RS1; \
  movq RS1, TEMP1; \
  cmp $CKB_VM_ASM_RISCV_MAX_MEMORY, TEMP1; \                                     <----------------- 1st
  jae .exit_out_of_bound; \
  addq length, TEMP1; \
  cmp $CKB_VM_ASM_RISCV_MAX_MEMORY, TEMP1; \                                     <----------------- 2nd
  ja .exit_out_of_bound

#define CHECK_READ(address_reg, length) \
  movq address_reg, TEMP1; \
  shr $CKB_VM_ASM_MEMORY_FRAME_SHIFTS, TEMP1; \
  cmp $CKB_VM_ASM_MEMORY_FRAMES, TEMP1; \                                        <----------------- 3rd
  jae .exit_out_of_bound; \
  movzbl CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_FRAMES(MACHINE, TEMP1), TEMP2d; \
  cmp $0, TEMP2d; \
  jne 1f; \
  movb $1, CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_FRAMES(MACHINE, TEMP1); \
  PREPCALL; \
  MOV_TEMP1_TO_ARG1; \
  MOV_MACHINE_TO_ARG2; \
  CALL_INITED_MEMORY; \
  POSTCALL; \
1: \
  movq address_reg, TEMP1; \
  addq $length, TEMP1; \
  subq $1, TEMP1; \
  shr $CKB_VM_ASM_MEMORY_FRAME_SHIFTS, TEMP1; \                                  <----------------- 4th
  cmp $CKB_VM_ASM_MEMORY_FRAMES, TEMP1; \
  jae .exit_out_of_bound; \
  movzbl CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_FRAMES(MACHINE, TEMP1), TEMP2d; \
  cmp $0, TEMP2d; \
  jne 2f; \
  movb $1, CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_FRAMES(MACHINE, TEMP1);\
[...]

PR #303, opened 2022-11-03
Remove duplicate out of bounds checks

mohanson:
#302


xxuejie:
It's not directly related to the changes made by this PR, but there's something that stands out after we change this PR:

* For load operations of version 0, we manually copy the instructions used to check bounds, like [this](https://github.com/nervosnetwork/ckb-vm/blob/f66f147c52eb154f388f5bdc02f7cd1ebba3985f/src/machine/asm/execute_x64.S#L664-L671)
* But for load operations of version 1, a macro `CHECK_READ_BOUND_VERSION1` is used instead

There is also hidden knowledge that a `CHECK_READ` macro must be paired with either `CHECK_READ_BOUND_VERSION1`, or a series of manually copied instructions. All of those can make future maintenance a nuanced job. Which leads to the following suggestion:

* Add a new `CHECK_READ_BOUND_VERSION0` which resembles `CHECK_READ_BOUND_VERSION1`, but use logic from version 0.
* Rename `CHECK_READ` to `CHECK_READ_PERMISSIONS`

I believe we will thank ourselves in the future by doing those changes.

mohanson:
> It's not directly related to the changes made by this PR, but there's something that stands out after we change this PR:
> 
> * For load operations of version 0, we manually copy the instructions used to check bounds, like [this](https://github.com/nervosnetwork/ckb-vm/blob/f66f147c52eb154f388f5bdc02f7cd1ebba3985f/src/machine/asm/execute_x64.S#L664-L671)
> * But for load operations of version 1, a macro `CHECK_READ_BOUND_VERSION1` is used instead
> 
> There is also hidden knowledge that a `CHECK_READ` macro must be paired with either `CHECK_READ_BOUND_VERSION1`, or a series of manually copied instructions. All of those can make future maintenance a nuanced job. Which leads to the following suggestion:
> 
> * Add a new `CHECK_READ_BOUND_VERSION0` which resembles `CHECK_READ_BOUND_VERSION1`, but use logic from version 0.
> * Rename `CHECK_READ` to `CHECK_READ_PERMISSIONS`
> 
> I believe we will thank ourselves in the future by doing those changes.

The code has been modified according to your comments

xxuejie (approved):

PR #305, opened 2022-11-09
Fix Asm aarch64 bugs

joii2020:
1. RS2 occupies x18 register
2. Version is u32, but read out of bounds


mohanson:
The documentation explains why `x18` cannot be used

https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms

> The platforms reserve register x18. Don’t use this register.

mohanson (dismissed):
LGTM

src/machine/asm/execute_aarch64.S
```diff
@@ -41,20 +41,22 @@
...
 
 #define PREPCALL \
-  stp x0, x8, [sp, -96]! SEP \
+  stp x0, x8, [sp, -112]! SEP \
   stp x9, x10, [sp, 16] SEP \
   stp x11, x12, [sp, 32] SEP \
   stp x13, x14, [sp, 48] SEP \
   stp x15, x16, [sp, 64] SEP \
-  stp x17, x18, [sp, 80]
+  stp x17, x18, [sp, 80] SEP \
```
xxuejie: If we are not using x18, there's no need to store x18 on the stack. We can simply replace x18 here with x24, which means there's no need to store x25 as well.

src/machine/asm/execute_aarch64.S
```diff
@@ -23,8 +23,8 @@
 #define RD x16
 #define RS1 x17
 #define RS1w w17
-#define RS2 x18
-#define RS2w w18
+#define RS2 x24
```
xxuejie, line 26: The aarch64 version is nicely maintained so actual registers can easily be changed, can we re-assign RS2 as x19 and adjust all registers below accordingly? Having a sequential order can be nicer when debugging assembly code.

src/machine/asm/execute_aarch64.S
```diff
@@ -242,9 +244,10 @@ _ckb_vm_x64_execute:
 .globl ckb_vm_x64_execute
 ckb_vm_x64_execute:
 #endif
-  stp x19, x20, [sp, -48]!
+  stp x19, x20, [sp, -64]!
   stp x21, x22, [sp, 16]
   stp x23, x30, [sp, 32]
+  stp x24, x25, [sp, 48]
```
xxuejie: Notice we are not using x30 in this file at all, can we replace x30 with x24 and see if it works?

joii2020: Yes is works. We can using ohter registers to store some memory data.

mohanson (dismissed):

mohanson:
cc @xxuejie

xxuejie:
While not exactly this PR's problem, I do recommend that we revise the register used here, per [A64 calling conventions](https://en.wikipedia.org/wiki/Calling_convention#ARM_(A64)), right now we are still using x8, x16 and x17, this might bring future problems. I would suggest the following layouts:

* MACHINE: x0
* TEMP1: x9
* TEMP2: x10
* TEMP3: x11
* TEMP4: x12
* TEMP5: x13
* TRACE: x19
* INST_PC: x20
* INST_ARGS: x21
* RD: x22
* RS1: x23
* RS2: x24
* RS3: x25
* IMMEDIATE: x26
* FLAG: x27
* REGISTER_BASE: x28

Notice I've used more callee-saved registers so as to minimize load/store needed for PREPCALL/POSTCALL, hoping it can reduce the overall code size a bit.

mohanson:
> Notice we are not using x30 in this file at all, can we replace x30 with x24 and see if it works?

I tried it myself, if I remove the x30, the tests failed with `signal: 11, SIGSEGV: invalid memory reference`

xxuejie (approved):

PR #308, opened 2022-11-28
Add load_bytes interface for memory

joii2020:
Add load_bytes interface for trait Memory

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -190,6 +190,22 @@ fn check_memory_inited(
     Ok(())
 }
 
+fn check_memory_block_inited(
+    machine: &mut Box<AsmCoreMachine>,
+    addr: u64,
+    size: usize,
+) -> Result<(), Error> {
+    let begin_page = addr >> RISCV_PAGE_SHIFTS;
+    let end_page = (addr + size as u64) >> RISCV_PAGE_SHIFTS;
+    for page in begin_page..(end_page + 1) {
+        if page as usize >= RISCV_PAGES {
+            return Err(Error::MemOutOfBound);
+        }
+        check_memory(machine, page);
+    }
+    Ok(())
+}
+
 impl Memory for Box<AsmCoreMachine> {
     type REG = u64;
 
@@ -299,13 +315,7 @@ impl Memory for Box<AsmCoreMachine> {
     }
 
     fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error> {
-        if length == 0 {
-            return Ok(Bytes::new());
-        }
-        let page_indices = get_page_indices(addr, length as u64)?;
-        for page in page_indices.0..=page_indices.1 {
-            check_memory(self, page);
-        }
+        check_memory_block_inited(self, addr, length)?;
         Ok(Bytes::from(
             self.memory[(addr as usize)..(addr as usize + length)].to_vec(),
         ))
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -315,6 +315,9 @@ impl Memory for Box<AsmCoreMachine> {
     }
 
     fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error> {
+        if length == 0 {
+            return Ok(Bytes::new());
+        }
         check_memory_block_inited(self, addr, length)?;
         Ok(Bytes::from(
             self.memory[(addr as usize)..(addr as usize + length)].to_vec(),
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -190,22 +190,6 @@ fn check_memory_inited(
     Ok(())
 }
 
-fn check_memory_block_inited(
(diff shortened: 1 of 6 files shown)


src/machine/asm/mod.rs
```diff
@@ -298,6 +298,18 @@ impl Memory for Box<AsmCoreMachine> {
         Ok(())
     }
 
+    fn load_bytes(&mut self, addr: u64, out_value: &mut [u8]) -> Result<(), Error> {
```
mohanson: This function signature is too C-style, although there is nothing wrong with it, but I prefer 

```
fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error>
```

src/memory/flat.rs
```diff
@@ -203,4 +203,15 @@ impl<R: Register> Memory for FlatMemory<R> {
...
     }
+
+    fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error> {
+        if length == 0 {
+            return Ok(Bytes::new());
+        }
+        let page_indices = get_page_indices(addr.to_u64(), length as u64)?;
+        set_dirty(self, &page_indices)?;
+        Ok(Bytes::from(
+            self[addr as usize..(addr + length as u64) as usize].to_vec(),
```
mohanson: `addr as usize + length`

src/memory/sparse.rs
```diff
@@ -193,6 +193,32 @@ impl<R: Register> Memory for SparseMemory<R> {
...
 
+    fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error> {
+        if length == 0 {
+            return Ok(Bytes::new());
+        }
+        let mut current_page_addr = round_page_down(addr);
+        let mut current_page_offset = addr - current_page_addr;
+        let mut need_read_len = length;
+        let mut out_value = Vec::new();
+        out_value.resize(length, 0);
```
mohanson: I don't recommend pre-allocating memory for vec here:

In the design of syscall set_content, the value of length comes from the user, so we have to consider a question, if the user enters a wrong length, how should ckb-vm respond.

In your code, if length is a large number, ckb-vm will oom, with disastrous results.

mohanson: Therefore I think it is better to use Vec::extend to dynamically add memory data

src/memory/flat.rs
```diff
@@ -203,4 +203,15 @@ impl<R: Register> Memory for FlatMemory<R> {
         memset(&mut self[addr as usize..(addr + size) as usize], value);
         Ok(())
     }
+
+    fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error> {
+        if length == 0 {
+            return Ok(Bytes::new());
+        }
+        let page_indices = get_page_indices(addr.to_u64(), length as u64)?;
+        set_dirty(self, &page_indices)?;
```
mohanson: Don't need to `set_dirty` when loading data

src/machine/asm/mod.rs
```diff
@@ -298,6 +298,19 @@ impl Memory for Box<AsmCoreMachine> {
         Ok(())
     }
 
+    fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error> {
+        if length == 0 {
+            return Ok(Bytes::new());
+        }
```
mohanson, line 304: Must check Error::MemOutOfBound. See `load64`

src/memory/flat.rs
```diff
@@ -203,4 +203,15 @@ impl<R: Register> Memory for FlatMemory<R> {
         memset(&mut self[addr as usize..(addr + size) as usize], value);
         Ok(())
     }
+
+    fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error> {
+        if length == 0 {
+            return Ok(Bytes::new());
+        }
```
mohanson, line 210: Must check Error::MemOutOfBound.

src/memory/wxorx.rs
```diff
@@ -140,4 +140,13 @@ impl<M: Memory> Memory for WXorXMemory<M> {
         check_permission(self, &page_indices, FLAG_WRITABLE)?;
         self.inner.store_byte(addr, size, value)
     }
+
+    fn load_bytes(&mut self, addr: u64, length: usize) -> Result<Bytes, Error> {
+        if length == 0 {
+            return Ok(Bytes::new());
+        }
+        let page_indices = get_page_indices(addr, length as u64)?;
+        check_permission(self, &page_indices, FLAG_WRITABLE)?;
```
mohanson: Why check FLAG_WRITABLE when **loading** data?

src/memory/flat.rs
```diff
@@ -203,4 +203,16 @@ impl<R: Register> Memory for FlatMemory<R> {
...
+        if length + addr as usize > self.memory_size() {
```
mohanson: `addr` and `length` may be a dynamic value given by script user, so you must check length + addr is not overflowed. See how `load64()` does it.

mohanson (approved):

PR #311, opened 2022-12-07
chore: remove unnecessary instructions in trace_end

quake:
In the previous code before https://github.com/nervosnetwork/ckb-vm/pull/298 , `CKB_VM_ASM_LABEL_OP_CUSTOM_TRACE_END` and `prepare_trace` were sharing the same code block. 

Now the `CKB_VM_ASM_LABEL_OP_CUSTOM_TRACE_END` will only be triggered when the next trace is executed sequentially, it does not run at address 0, so we can remove the unnecessary length checking here.

### DIFF
--- a/src/machine/asm/execute_aarch64.S
+++ b/src/machine/asm/execute_aarch64.S
@@ -266,8 +266,6 @@ ckb_vm_x64_execute:
   cmp TEMP4, TEMP3
   bne .exit_trace
   ldrb TEMP3w, [TRACE, CKB_VM_ASM_TRACE_OFFSET_LENGTH]
-  cmp TEMP3w, 0
-  beq .exit_trace
   ldr TEMP2, [MACHINE, CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_CYCLES]
   ldr TEMP1, [TRACE, CKB_VM_ASM_TRACE_OFFSET_CYCLES]
   add TEMP2, TEMP2, TEMP1
--- a/src/machine/asm/execute_x64.S
+++ b/src/machine/asm/execute_x64.S
@@ -368,8 +368,6 @@ ckb_vm_x64_execute:
   cmp %rcx, %rdx
   jne .exit_trace
   movzbl CKB_VM_ASM_TRACE_OFFSET_LENGTH(TRACE), %edx
-  cmp $0, %rdx
-  je .exit_trace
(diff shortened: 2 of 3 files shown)


mohanson:
Consider a testcase

**main.S**
```
  .global _start
_start:
  li a0, 0
  li a7, 93
  ecall
```

**main.lds**
```
SECTIONS
{
. = 0x00;
.text : { *(.text) }
}
```

```sh
$ riscv64-unknown-elf-as -o main.o main.S 
$ riscv64-unknown-elf-ld -T main.lds -o main main.o 

$ ckb-vm-runner /tmp/main
Segmentation fault (core dumped)
```

quake:
I didn't realize that the instruction could be constructed in the zero address, close this PR and add a test case PR #312

xxuejie:
This is not necessarily a problem, but I do have something that I want to raise a discussion here:

Depending on how you look at the problem, the reason that the about test case triggers a segfault for this PR, could be a coincidence: `.CKB_VM_ASM_LABEL_OP_CUSTOM_TRACE_END` occurs before `.prepare_trace` in this [part](https://github.com/nervosnetwork/ckb-vm/blob/66df56922eb7842fcbd5d94c89f7aa7f133d8320/src/machine/asm/execute_x64.S#L347-L413) of code. In other words, after the initial setup phase of `ckb_vm_x64_execute` function, we first execute `.CKB_VM_ASM_LABEL_OP_CUSTOM_TRACE_END`, not `.prepare_trace`.

If we swap the code order between `.CKB_VM_ASM_LABEL_OP_CUSTOM_TRACE_END` and `.prepare_trace`, I believe the above test case would work.

Right now, the 2 labels above are really used for 2 different purposes:

* `.CKB_VM_ASM_LABEL_OP_CUSTOM_TRACE_END` is used to handle sequential code execution
* `.prepare_trace` is used to handle jumps

That brings a question: when we first execute `ckb_vm_x64_execute`, are we expecting to handle a jump, or a sequential code execution? I tend to believe that we should be expecting to handle jumps when we first enter `ckb_vm_x64_execute`, which means we should execute `.prepare_trace` after the initial setup in `ckb_vm_x64_execute`. That also means the code in this PR could be applied.

Thoughts?

quake:
+1 for switching the label, ~~btw, if I understand correctly, the cycle overflow and max cycles exceeded checking can be removed from `. CKB_VM_ASM_LABEL_OP_CUSTOM_TRACE_END` also after switching?~~

quake:
will submit a new pr later

PR #316, opened 2023-01-16
Assembly VM optimizations

xxuejie:
Please see individual commits for details on the optimizations


mohanson:
The decoder for compressed instructions should also produce two versions of the instruction.

xxuejie:
> The decoder for compressed instructions should also produce two versions of the instruction.

I actually thought about this, but 2 things came up:

1. Will we use mop together with version 0 of the code? ckb does not use such a configuration, other use cases should stick to latest version forever.
2. If you think further, fusing should indeed only happen with version 1's JALR, since version 0's JALR has different behavior. In other words, mop would generate different results for JALR in version 0, comparing to non-mop version.

mohanson:
> > The decoder for compressed instructions should also produce two versions of the instruction.
> 
> I actually thought about this, but 2 things came up:
> 
> 1. Will we use mop together with version 0 of the code? ckb does not use such a configuration, other use cases should stick to latest version forever.
> 2. If you think further, fusing should indeed only happen with version 1's JALR, since version 0's JALR has different behavior. In other words, mop would generate different results for JALR in version 0, comparing to non-mop version.

I mean the `C` in `IMC`

mohanson:
My mistake, you have fixed it

mohanson (approved):
Looks good to me.

PR #319, opened 2023-02-02
Add comments for macro-op ADC/SBB

mohanson:


xxuejie:
Can we organize a doc maybe into this repository, other than using a github issue?

xxuejie:
Sorry I actually mean if we can create a `docs` folder, then create a `mops.md` file in that folder, in which we can document all the mops supported by ckb-vm.

mohanson:
ok, let me deal with it.

PR #320, opened 2023-02-08
perf: Small decoder optimizations:

xxuejie:
* Use an instruction cache key that reduces actual decoding times
* Cache mop instructions as well as raw instructions. Though this put both instruction types in the same cache, I don't think this will cause any problems.

The goal here is to boost performance gained by mop for larger contracts, while keeping negligible performance difference between mop and non-mop version for secp256k1. Per my test, this would boost the performance running halo2, while keeping <0.05ms runtime difference for secp256k1 benchmarks.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -90,8 +90,7 @@ impl Decoder {
         if pc as usize >= RISCV_MAX_MEMORY {
             return Err(Error::MemOutOfBound);
         }
-        // according to RISC-V instruction encoding, the lowest bit in PC will always be zero
-        let instruction_cache_key = (pc >> 1) as usize % INSTRUCTION_CACHE_SIZE;
+        let instruction_cache_key = cache_key(pc);
         let cached_instruction = self.instructions_cache[instruction_cache_key];
         if cached_instruction.0 == pc {
             return Ok(cached_instruction.1);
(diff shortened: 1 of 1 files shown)


mohanson:
```
((pc & 0xFF) | (pc >> 12 << 8)) as usize % INSTRUCTION_CACHE_SIZE
```

This optimization looks clever, how did you come up with it?

xxuejie:
> ```
> ((pc & 0xFF) | (pc >> 12 << 8)) as usize % INSTRUCTION_CACHE_SIZE
> ```
> 
> This optimization looks clever, how did you come up with it?

The original inspiration comes from a manual trace of cache eviction, I notice a pattern that the code in memcpy is decoded many times, then I noticed a frequent jump to memcpy and back. It reminds me of [bochs paper](https://bochs.sourceforge.io/Virtualization_Without_Hardware_Final.pdf) I read before, which also uses a trace cache index formula involving both higher address and lower address.

After that, it's just many different tries till I found out that 12 and 8 gives the best result on both halo2 and secp256k1.

I actually leave a comment explaining the rationale, and that the number comes from "empirical evidence".

mohanson (approved):

xxuejie:
Also, it's worthwhile to look into similar optimization on trace key calculation in ckb, which is on my list next.

xxuejie:
It does seem this will affect more than we originally thought, I will close this PR and create a new one that only includes the cache key calculation part.

PR #321, opened 2023-02-08
feat: Introduce new mops for version 2

xxuejie:
This change introduces 5 new mops for ckb-vm version 2. Those new mops all come from a direct observation of generated code from modern LLVM/GCC compilers. While they have less impact on secp256k1 contract(1183580 -> 1148321), they have more significant impact on complicated contracts. Using halo2 as an example:

* The cycle consumption is reduced from 77309063 to 67521505
* Actual runtime is reduced from 129ms to 111ms


src/machine/asm/execute_aarch64.S
```diff
@@ -1594,6 +1611,66 @@ ckb_vm_x64_execute:
...
+  WRITE_RS3(TEMP1)
+  NEXT_INST
+.CKB_VM_ASM_LABEL_OP_SBBS:
+  DECODE_R4
+  mov TEMP1, 0
+  ldr TEMP3, REGISTER_ADDRESS(RS1)
+  ldr TEMP4, REGISTER_ADDRESS(RS2)
+  subs TEMP3, TEMP3, TEMP4
+  adc TEMP1, TEMP1, TEMP1
+  eor TEMP1, TEMP1, 1
```
mohanson, line 1631: Why xored TEMP1?

xxuejie: arm64 uses different borrow logic in subtraction here: when a borrow happened, the carry bit is 0, otherwise the carry bit is 1. This is the opposite of what we want, hence the xor.

mohanson: Understand, it is said to simplify chip design

https://stackoverflow.com/questions/53065579/confusion-about-arm-documentation-on-carry-flag

tests/programs/mop_jump_rel_version1_bug.S
```diff
@@ -0,0 +1,12 @@
...
+_start:
+  auipc ra, 0x80000
+  jalr ra, ra, -0x7ff
+  li a0, 1
+  li a7, 93
+  ecall
+1:
+  li a0, 0
+  li a7, 93
+  ecall
```
mohanson: These lines are useless, please delete them.

xxuejie: Done

mohanson (approved):

PR #327, opened 2023-02-21
feat(cost_model): add default instruction cycle func

mohanson:
ckb-vm will maintain a default `instruction_cycle_func`. This can simplify the development of some small tools based on ckb-vm.

### DIFF
--- a/src/cost_model.rs
+++ b/src/cost_model.rs
@@ -1,0 +1,70 @@
+use crate::{
+    instructions::{extract_opcode, insts},
+    Instruction,
+};
+
+// Returns the spent cycles to execute the secific instruction.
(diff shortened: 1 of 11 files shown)


XuJiandong (approved):

xxuejie:
There seems to be many changes that break a big `use` into multiple smaller ones, is this due to new behaviors of rust-fmt, or a delibrate choice?

mohanson:
> There seems to be many changes that break a big `use` into multiple smaller ones, is this due to new behaviors of rust-fmt, or a delibrate choice?

It's my own peculiar inclination, it makes the code more compact.

xxuejie (approved):

PR #329, opened 2023-02-28
Implement atomic instructions(A Standard Extension)

mohanson:
- Update CI with A Extension testsuite
- `ckb_vm::run` and `examples/ckb-vm-runner` adds ISA_A by default


src/machine/mod.rs
```diff
@@ -44,6 +44,10 @@ pub trait CoreMachine {
     fn registers(&self) -> &[Self::REG];
     fn set_register(&mut self, idx: usize, value: Self::REG);
 
+    // Load reservation address for atomic extension.
+    fn lr(&self) -> &Self::REG;
```
xxuejie: Several things:

* `lr` and `set_lr` should be components of the memory trait, they should not belong to `CoreMachine`
* Notice depending on the actual architecture, there might be more than one reserved addresses at the same time. In our Rust implementation, we should make an array of `load_reservation_address`, in the assembly code, we definitely want to limit the reservations available at the same time, I'm thinking something like 2 but it's worth checking generated code from popular library to see how they would generate code here.

src/machine/asm/execute_x64.S
```diff
@@ -1251,6 +1254,260 @@ ckb_vm_x64_execute:
...
+  NEXT_INST
+.sc_d_branch:
+  WRITE_RD_VALUE($1, TEMP1)
+  NEXT_INST
+.p2align 3
+.CKB_VM_ASM_LABEL_OP_AMOSWAP_D:
+  DECODE_R
+  movq REGISTER_ADDRESS(RS1), RS1
+  movq REGISTER_ADDRESS(RS2r), RS2r
+  CHECK_READ_VERSION1(RS1, 8)
```
xxuejie: For the atomic operations, we will need to check for memory write permissions as well

src/machine/mod.rs
xxuejie: One way to deal with this, could be that we can treat `LR` and `SC` as slowpath instructions, hence reducing the need to implement complicated logic in assembly.

mohanson: 1) It makes sense to move lr and set_lr to the memory trait.

2) In the implementation of lr and sc, I refer to the code of spike, It only allows one load_reservation_address to exist: https://github.com/riscv-software-src/riscv-isa-sim/blob/8cabd4a29b183973078da0f7766b4d5f2005eeba/riscv/mmu.h#L308

xxuejie: You are right, it does seem spike is also simplifying code here, in that sense let's stick to 1 `load_resevation_address` as well.

src/machine/asm/execute_x64.S
mohanson: Fixed

src/machine/asm/execute_x64.S
```diff
@@ -1251,6 +1254,278 @@ ckb_vm_x64_execute:
...
+  NEXT_INST
+.sc_w_branch:
+  WRITE_RD_VALUE($1, TEMP1)
+  NEXT_INST
+.p2align 3
+.CKB_VM_ASM_LABEL_OP_AMOSWAP_W:
+  DECODE_R
+  movq REGISTER_ADDRESS(RS1), RS1
+  movq REGISTER_ADDRESS(RS2r), RS2r
+  CHECK_READ_VERSION1(RS1, 4)
```
xxuejie: I think we can skip `CHECK_READ_VERSION1`, when we have `CHECK_WRITE` below, correct?

mohanson: It is necessary due to the lazy initialization of memory.

xxuejie: I think now `CHECK_WRITE` does do lazy memory initialization: https://github.com/nervosnetwork/ckb-vm/blob/1e600830cd9555bbdee19c9cd373065873362c09/src/machine/asm/execute_x64.S#L291, or am I missing something?

mohanson: Yes, you are right

src/machine/asm/execute_x64.S
```diff
@@ -1251,6 +1254,260 @@ ckb_vm_x64_execute:
...
+.p2align 3
+.CKB_VM_ASM_LABEL_OP_SC_W:
+  DECODE_R
+  movq REGISTER_ADDRESS(RS1), RS1
+  movq REGISTER_ADDRESS(RS2r), RS2r
+  CHECK_WRITE(RS1, TEMP1d, 4)
+  movq LOAD_RESERVATION_ADDRESS, TEMP1
+  cmp RS1, TEMP1
+  jne .sc_w_branch
+  movq RS2r, CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_MEMORY(MACHINE, RS1)
```
xxuejie: Should we use `RS2rd` here?

src/machine/asm/execute_x64.S
```diff
@@ -1251,6 +1254,260 @@ ckb_vm_x64_execute:
...
+  movq $UINT64_MAX, LOAD_RESERVATION_ADDRESS
+  WRITE_RD_VALUE($0, TEMP1)
+  NEXT_INST
+.sc_w_branch:
```
xxuejie: Per RISC-V's spec: in case of failure, reservation should also be cleared.

src/instructions/execute.rs
```diff
@@ -164,6 +165,219 @@ pub fn execute_instruction<Mac: Machine>(
...
+            let i = Rtype(inst);
+            let address = machine.registers()[i.rs1() as usize].clone();
+            let condition = address.eq(machine.memory().lr());
+            let mem_value = condition.cond(
+                &machine.registers()[i.rs2() as usize].clone(),
+                &machine.memory_mut().load32(&address)?,
+            );
+            let rd_value = condition.cond(&Mac::REG::from_u8(0), &Mac::REG::from_u8(1));
+            // Make sure that sc-after-successful-sc fails.
+            let lr_value = condition.cond(&Mac::REG::from_u64(u64::MAX), &address);
```
xxuejie: `lr` should also be cleared in case of failure per RISC-V's spec

src/machine/asm/execute_x64.S
```diff
@@ -1251,6 +1254,260 @@ ckb_vm_x64_execute:
...
+  NEXT_INST
+.p2align 3
+.CKB_VM_ASM_LABEL_OP_AMOADD_W:
+  DECODE_R
+  movq REGISTER_ADDRESS(RS1), RS1
+  movq REGISTER_ADDRESS(RS2r), RS2r
+  CHECK_WRITE(RS1, TEMP1d, 4)
+  movslq CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_MEMORY(MACHINE, RS1), TEMP1
+  WRITE_RD(TEMP1)
+  add TEMP1, RS2r
```
xxuejie, line 1298: I noticed a line in RISC-V's spec:

`For RV64, 32-bit AMOs always sign-extend the value placed in rd, and ignore the upper 32 bits of the original value of rs2`

Does this mean we need to clear the upper 32 bits of `rs2` here for all `AMO*.W` operations?

src/machine/asm/execute_x64.S
mohanson: Fixed.

src/machine/asm/execute_x64.S
mohanson: After `add TEMP1, RS2r`, we write the result to memory using the following statement:

```
mov RS2rd, CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_MEMORY(MACHINE, RS1)
```

It has no affects the final result whether or not the upper 32 bits of rs2 are cleared.

xxuejie: 2 things:

* The spec requires `rd` to be sign-extended, which means `mov RS2rd` might not be correct here?
* For `add` you are correct, this might end up in the same result, but what about `MIN`/`MAX`? Ignoring upper 32 bits or sign-extending can lead to different result, correct?

xxuejie:
One more thing: it seems the new issues we found here are not covered by `riscv-tests`, any chance we can add some unit tests covering checks on those behavior?

src/machine/asm/execute_x64.S
mohanson: - `mov RS2rd, CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_MEMORY(MACHINE, RS1)` moves RS2rd to memory, not `rd`, and rd = sign_ext(memory[rs1])
- Good question, I need to take a closer look.

mohanson:
> One more thing: it seems the new issues we found here are not covered by `riscv-tests`, any chance we can add some unit tests covering checks on those behavior?

Yes, I'm going to create some PRs to riscv-tests, but before that I will add tests for ckb-vm.

xxuejie (approved):

PR #333, opened 2023-03-30
Fuzz isa_a with spike-ffi

mohanson:
- Implement spike-ffi
- Fuzz A extension between interpreter, asm and spike.

```sh
$ cd fuzz_spike_ffi
$ sh build.sh
$ cd ..

$ cargo +nightly fuzz run isa_a
```


fuzz_spike_ffi/build.sh
```diff
@@ -0,0 +1,22 @@
+set -ex
```
XuJiandong: Add `readme` about how to build fuzz_spike_ffi crate and how it works.

XuJiandong (changes requested):

xxuejie:
IMHO it might be better to put `fuzz_spike_ffi` in `ckb-vm-test-suite`

mohanson:
I created a crate so that ckb-vm only needs to import it into dependency

source: https://github.com/nervosnetwork/ckb-vm-test-suite/tree/master/spike-sys
crates.io https://crates.io/crates/spike-sys

cc @XuJiandong @xxuejie

XuJiandong (dismissed):

fuzz/Cargo.toml
```diff
@@ -9,16 +9,19 @@ edition = "2018"
...
+[dependencies.ckb-vm]
+path = ".."
+features = ["asm"]
 
 # Prevent this from interfering with workspaces
 [workspace]
 members = ["."]
 
 [[bin]]
-name = "asm"
```
xxuejie: Why do we replace one with the other, instead of keeping both fuzzing targets? I thought they can work for different use cases.

mohanson: I added it back to the fuzz targets. It tests the load_program function which I overlooked earlier.

xxuejie (approved):

PR #335, opened 2023-04-10
Add probes for syscalls and functions

contrun:
There are two probes added here. One is used to track syscall parameters and return values. One is used to track function calls and returns. The functionality of the code here has been manually verified with https://github.com/contrun/ckb-vm-bpf-toolkit/blob/731ce2095eb4074e57616dad538ed51bf598f255/tools/uretprobe.py. I used https://github.com/contrun/ckb-standalone-debugger/tree/use-ckb-vm-with-probe-support (which requires https://github.com/contrun/ckb-vm/tree/add-probes-for-syscalls-and-functions that have the same code change to this commit except that is based on v0.22.2) to show the usage of these probes. For example, below is a sample output of running `sudo env RUST_LOG=debug ./tools/memdump.py --bin ../ckb-lua/build/lua-loader.debug -- -e 'print(42)' --bpf-func "^test_func$" --bpf-regs a0 --bpf-debugger ../ckb-standalone-debugger/target/release/ckb-debugger`

```
Func test_func has been jumped to/from 2 times!
Func test_func has been called 1 times!
Func test_func has returned 1 times!
Dumping return value counts for func test_func
return value: 000000000005d020, count: 1
Dumping meomry addresses for func test_func
memory addr: 00007f1474126010, content: 0000000000000001
Dumping meomry content for func test_func
memory addr: 000000000005d020, content: 31302f2e2d2c2b2a
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ enable-chaos-mode-by-default = ["ckb-vm-definitions/enable-chaos-mode-by-default
 # Disable slow tests to run miri on CI
 miri-ci = []
 pprof = []
+probes = ["dep:probe"]
 
 [dependencies]
 byteorder = "1"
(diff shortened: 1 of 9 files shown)


mohanson:
It is better to place it in the [ckb-standalone-deugger](https://github.com/nervosnetwork/ckb-standalone-debugger).

PR #337, opened 2023-04-27
Define the cycles consumption of A instruction set

mohanson:
Rules:
1) 32-bit memory reading and writing consumes 3 cycles, 64-bit memory reading and writing consumes 2 cycles.
2) All A instructions have a base cost of 1 cycles.


XuJiandong:
Add some description into code as comments about how to calculate the final cycles.

XuJiandong (dismissed):

src/cost_model.rs
```diff
@@ -57,6 +57,29 @@ pub fn estimate_cycles(i: Instruction) -> u64 {
...
+        insts::OP_AMOXOR_W => 7,
+        insts::OP_AMOAND_W => 7,
+        insts::OP_AMOOR_W => 7,
+        insts::OP_AMOMIN_W => 7,
+        insts::OP_AMOMAX_W => 7,
+        insts::OP_AMOMINU_W => 7,
+        insts::OP_AMOMAXU_W => 7,
+        insts::OP_LR_D => 3,
+        insts::OP_SC_D => 3,
+        insts::OP_AMOSWAP_D => 5,
```
xxuejie: I was slightly confused, how is this 5 calculated?

mohanson: OP_AMOSWAP_D contains a memory load(2 cycles) and a memory store(2 cycles), and the swap operation takes 1 cycles

xxuejie: Okay I see this now but honestly I have slightly different opinions. In CPU's perspective, each of the atomic operations essentially only *touch* one memory location, yes it is one read and one write but the time that matters, is the time to load the memory location into CPU cache, which happens only once here. Rest work is simply touching cache and can basically be counted as free. So in my opinions, this should be 1 cycle for calculation + 2 cycles for memory operations, which is 3 cycles.

mohanson: This point of view makes sense.

xxuejie (approved):

PR #343, opened 2023-05-26
refactor: Remove initializers from Memory trait

xxuejie:
Different memory trait impls might require different ways of initializing,what's more, having initializers in Memory trait would prevent us from writing `Box<dyn Memory<u64>>`. What we really need here, is a reset memory function. This changes uses `reset_memory` to replace initializers in Memory trait.


mohanson (approved):
LGTM

PR #345, opened 2023-06-02
feat: Add alternative snapshot design.

xxuejie:
This change adds a different snapshot design, that takes advantage of a newly defined trait named `DataSource`. DataSource is modeled after environmental data that would stay the same and most likely immutable for the entire duration of a CKB-VM program's lifecycle. An example could be the enclosing transaction (`ResolvedTransaction` in CKB's term) for CKB's smart contracts. No matter where and when we are running the smart contract, the enclosing transaction is always the same. By exploiting this property, we can reduce the size of a snapshot greatly: when a piece of data already exist in the DataSource, we won't need to store the same data in the snapshot. As long as we maintain enough information to locate them from the DataSource, the snapshot can successfully be used to resume the full running state of a CKB-VM Machine.


mohanson (dismissed):

XuJiandong:
Let us pause: Is it necessary to include this particular feature in the upcoming hard fork?

xxuejie:
> Let us pause: Is it necessary to include this particular feature in the upcoming hard fork?

I think this is just an added feature, if you don't use it, it won't affect existing features. Even if you use it in the code, it really is just an implementation detail that does not change any behavior. To answer the question more directly: tt depends on the CKB team if they want to use this in CKB. Maybe they want to use it, maybe not.

mohanson:
Can you fix the conflict?

xxuejie:
I can but have we reached the decision to merge this to develop branch? Or are you merely meant to test this?

mohanson:
I'm about to merge it into the develop branch.

xxuejie:
> I'm about to merge it into the develop branch.

Already rebased.

mohanson (approved):

PR #347, opened 2023-06-06
perf: Group basic block end opcodes to eliminate slower pattern matching

xxuejie:


definitions/src/instructions.rs
```diff
@@ -86,158 +86,159 @@ macro_rules! __for_each_inst_inner {
...
+            (AUIPC, 0x9c),
+            (BEQ, 0x9d),
+            (BGE, 0x9e),
+            (BGEU, 0x9f),
+            (BLT, 0xa0),
+            (BLTU, 0xa1),
+            (BNE, 0xa2),
+            (EBREAK, 0xa3),
+            (ECALL, 0xa4),
+            (FENCE, 0xa5),
```
mohanson, line 235: Why FENCE should be a  branch instruction?

xxuejie: You are right they don't need to be branch instructions. The problem I want to solve, is that I believe it can benefit us from having some nop instructions that can manually **cut** instructions into traces. In other words, using `FENCE` here can manually mark the end of a trace, aiding better constructions of traces. This can unlock new potentials in the future IMHO. In reality, we won't see many occurences of fences as well, so I assume having them as branching instructions, won't affect current code.

mohanson (approved):

mohanson (approved):

PR #349, opened 2023-06-07
perf: Threadify Rust interpreter

xxuejie:
This change leverages a technique named "threaded interpreter" to speedup the Rust interpreter. Basically, it splits a giant match statement into multiple smaller individual functions, each handling a specific opcode. When we have a group of opcodes(e.g., a basic block), we can extract the handler function(also named "thread") for each instruction's opcode. Then we can simple run each handler function to execute each instructions. This way we can aid CPU's branch predictor to better predict what code to execute next.

Note this work is inspired from @mohanson's original work at here:

https://github.com/nervosnetwork/ckb-vm/commit/84223732f5e14f9b18db9a776c870d1bda38e33e

Reference:

* http://www.emulators.com/docs/nx25_nostradamus.htm

### DIFF
--- a/definitions/src/instructions.rs
+++ b/definitions/src/instructions.rs
@@ -48,26 +48,70 @@ pub use paste::paste;
 #[doc(hidden)]
 #[macro_export]
 macro_rules! __apply {
-    ($callback:ident, $(($name:ident, $code:expr)),*) => {
+    ((0, $callback:ident), $(($name:ident, $code:expr)),*) => {
         $crate::instructions::paste! {
             $(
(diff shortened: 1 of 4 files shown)


xxuejie:
Testing on a Ryzen 3900x, this change brings 10% performance boost to the Rust interpreter benchmarking secp256k1:

```
interpret secp256k1_bench
                        time:   [12.280 ms 12.288 ms 12.296 ms]
                        change: [-10.158% -10.034% -9.9061%] (p = 0.00 < 0.05)
                        Performance has improved.
```

mohanson (approved):

PR #351, opened 2023-06-08
perf: Skip permission check when initializing pages in AsmMachine

xxuejie:
Profiling data shows that when running secp256k1 benches, ~ 41% of the time to run `store_bytes` method for AsmCoreMachine, is spent in `check_memory` function. Upon further investigation, this is because right now we are always doing zero-initilization, even when a full page is overwritten from ELF data. This change adds a newtype pattern that provides fast `store_byte` / `store_bytes` that 1) bypasses permission checking since this is not required; 2) only initialize pages that partially contain ELF data. For pages that will be fully written with data from the ELF file, this skips zero-initialization completely.

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -190,6 +190,140 @@ fn check_memory_inited(
     Ok(())
 }
 
+// A newtype supporting fast store_byte / store_bytes without memory
+// permission checking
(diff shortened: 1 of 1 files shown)


xxuejie:
On a Ryzen 3900x machine, this change trims off ~0.1 ms runtime of a totally of previously ~2.7 ms runtime(~ 3%) to run secp256k1 benchmark.

mohanson (approved):

PR #354, opened 2023-07-06
feat!: generic syscall handlers

blckngm:
Syscall handlers with limited lifetime and are `!Send` can still be very useful.

We can support them (in addition to `'static + Send + Sync` handlers) by making the syscall handler type a generic type parameter. Unlike previous ckb-vm versions, we **are not adding a lifetime parameter to the machine types**.

I've also implemented `Syscall` for `()` (for when there's no syscalls) and `&mut impl SupportMachine -> Result` closures. `Vec<BoxedSyscalls>` is now handled by `impl Syscall for Vec<BoxedSyscalls>`.


blckngm:
See #355 instead.

PR #355, opened 2023-07-06
feat!: ExecutionContext = Syscall + Debug + InstructionCyclesFunc

blckngm:
* Use a type parameter instead of `Box<dyn>` to support different lifetimes, static/dynamic dispatching, Send/Sync or not.
* Use a single trait to reduce the number of type parameters needed. It's still possible to compose syscall, debugger and instruction cycles functions, see the builder methods or `ExecutionContextExt`.

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -1,7 +1,7 @@
 use ckb_vm_definitions::instructions::Instruction;
 
 use super::Error;
-use crate::{machine::SupportMachine, syscalls::Syscalls};
+use crate::{machine::SupportMachine, syscalls::Syscalls, Debugger};
 
 pub trait ExecutionContext<Mac: SupportMachine> {
     fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
@@ -82,25 +82,26 @@ where
     }
 }
 
-pub struct WithDebugger<Ctx, F> {
+pub struct WithDebugger<Ctx, Dbg> {
     pub(super) base: Ctx,
-    pub(super) debugger: F,
+    pub(super) debugger: Dbg,
 }
 
-impl<Ctx, F, Mac> ExecutionContext<Mac> for WithDebugger<Ctx, F>
+impl<Ctx, Dbg, Mac> ExecutionContext<Mac> for WithDebugger<Ctx, Dbg>
 where
     Mac: SupportMachine,
     Ctx: ExecutionContext<Mac>,
-    F: FnMut(&mut Mac) -> Result<(), Error>,
+    Dbg: Debugger<Mac>,
 {
     fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
-        self.base.initialize(machine)
+        self.base.initialize(machine)?;
+        self.debugger.initialize(machine)
     }
     fn ecall(&mut self, machine: &mut Mac) -> Result<bool, Error> {
         self.base.ecall(machine)
     }
     fn ebreak(&mut self, machine: &mut Mac) -> Result<(), Error> {
-        (self.debugger)(machine)
+        self.debugger.ebreak(machine)
     }
     fn instruction_cycles(&self, inst: Instruction) -> u64 {
         self.base.instruction_cycles(inst)
@@ -147,11 +148,10 @@ pub trait ExecutionContextExt<Mac: SupportMachine>: ExecutionContext<Mac> {
     }
 
     /// Replace the debugger.
-    fn with_debugger<F>(self, debugger: F) -> WithDebugger<Self, F>
+    fn with_debugger<Dbg>(self, debugger: Dbg) -> WithDebugger<Self, Dbg>
     where
         Self: Sized,
-        // For type inference.
-        F: FnMut(&mut Mac) -> Result<(), Error>,
+        Dbg: Debugger<Mac>,
     {
         WithDebugger {
             base: self,
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -5,8 +5,10 @@ use crate::{machine::SupportMachine, syscalls::Syscalls};
 
 pub trait ExecutionContext<Mac: SupportMachine> {
     fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
-        // We don't want to change param name to start with _ or others
-        // implementing this would need to remove the _.
+        // We don't want to change the param name to start with an
+        // underscore(_). It doesn't look good in docs. Also when someone
+        // implements this method with IDE completion, they will need to remove
+        // the staring underscore.
         #[allow(clippy::drop_ref)]
         drop(machine);
         Ok(())
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -9,26 +9,22 @@ pub trait ExecutionContext<Mac: SupportMachine> {
         // underscore(_). It doesn't look good in docs. Also when someone
         // implements this method with IDE completion, they will need to remove
         // the staring underscore.
-        #[allow(clippy::drop_ref)]
-        drop(machine);
+        let _ = machine;
         Ok(())
     }
     /// Return true if the syscall has been processed. If a module returns
     /// false, Machine would continue to leverage the next syscall module to
     /// process.
     fn ecall(&mut self, machine: &mut Mac) -> Result<bool, Error> {
-        #[allow(clippy::drop_ref)]
-        drop(machine);
+        let _ = machine;
         Ok(false)
     }
     fn ebreak(&mut self, machine: &mut Mac) -> Result<(), Error> {
-        #[allow(clippy::drop_ref)]
-        drop(machine);
+        let _ = machine;
         Ok(())
     }
     fn instruction_cycles(&self, inst: Instruction) -> u64 {
-        #[allow(clippy::drop_copy)]
-        drop(inst);
+        let _ = inst;
         0
     }
 }
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -1,0 +1,178 @@
+use ckb_vm_definitions::instructions::Instruction;
+
+use super::Error;
+use crate::{machine::SupportMachine, syscalls::Syscalls};
+
+pub trait ExecutionContext<Mac: SupportMachine> {
+    fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        // We don't want to change param name to start with _ or others
+        // implementing this would need to remove the _.
+        #[allow(clippy::drop_ref)]
+        drop(machine);
+        Ok(())
+    }
+    /// Return true if the syscall has been processed. If a module returns
+    /// false, Machine would continue to leverage the next syscall module to
+    /// process.
+    fn ecall(&mut self, machine: &mut Mac) -> Result<bool, Error> {
+        #[allow(clippy::drop_ref)]
+        drop(machine);
+        Ok(false)
+    }
+    fn ebreak(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        #[allow(clippy::drop_ref)]
+        drop(machine);
+        Ok(())
+    }
+    fn instruction_cycles(&self, inst: Instruction) -> u64 {
+        #[allow(clippy::drop_copy)]
+        drop(inst);
+        0
+    }
+}
+
+pub type BoxedExecutionContext<Mac> = Box<dyn ExecutionContext<Mac> + Send + Sync + 'static>;
+
+impl<Mac: SupportMachine> ExecutionContext<Mac> for BoxedExecutionContext<Mac> {
+    fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        (**self).initialize(machine)
+    }
+    fn ecall(&mut self, machine: &mut Mac) -> Result<bool, Error> {
+        (**self).ecall(machine)
+    }
+    fn ebreak(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        (**self).ebreak(machine)
+    }
+    fn instruction_cycles(&self, inst: Instruction) -> u64 {
+        (**self).instruction_cycles(inst)
+    }
+}
+
+impl<Mac: SupportMachine> ExecutionContext<Mac> for () {}
+
+pub struct WithSyscall<Ctx, Sys> {
+    pub(super) base: Ctx,
+    pub(super) syscall: Sys,
+}
+
+impl<Ctx, Sys, Mac> ExecutionContext<Mac> for WithSyscall<Ctx, Sys>
+where
+    Mac: SupportMachine,
+    Ctx: ExecutionContext<Mac>,
+    Sys: Syscalls<Mac>,
+{
+    fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        self.base.initialize(machine)?;
+        self.syscall.initialize(machine)
+    }
+    fn ecall(&mut self, machine: &mut Mac) -> Result<bool, Error> {
+        let processed = self.base.ecall(machine)?;
+        if processed {
+            return Ok(processed);
+        }
+        self.syscall.ecall(machine)
+    }
+    fn ebreak(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        self.base.ebreak(machine)
+    }
+    fn instruction_cycles(&self, inst: Instruction) -> u64 {
+        self.base.instruction_cycles(inst)
+    }
+}
+
+pub struct WithDebugger<Ctx, F> {
+    pub(super) base: Ctx,
+    pub(super) debugger: F,
+}
+
+impl<Ctx, F, Mac> ExecutionContext<Mac> for WithDebugger<Ctx, F>
+where
+    Mac: SupportMachine,
+    Ctx: ExecutionContext<Mac>,
+    F: FnMut(&mut Mac) -> Result<(), Error>,
+{
+    fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        self.base.initialize(machine)
+    }
+    fn ecall(&mut self, machine: &mut Mac) -> Result<bool, Error> {
+        self.base.ecall(machine)
+    }
+    fn ebreak(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        (self.debugger)(machine)
+    }
+    fn instruction_cycles(&self, inst: Instruction) -> u64 {
+        self.base.instruction_cycles(inst)
+    }
+}
+
+pub struct WithCyclesFunc<Ctx, F> {
+    pub(super) base: Ctx,
+    pub(super) cycles: F,
+}
+
+impl<Ctx, F, Mac> ExecutionContext<Mac> for WithCyclesFunc<Ctx, F>
+where
+    Mac: SupportMachine,
+    Ctx: ExecutionContext<Mac>,
+    F: Fn(Instruction) -> u64,
+{
+    fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        self.base.initialize(machine)
+    }
+    fn ecall(&mut self, machine: &mut Mac) -> Result<bool, Error> {
+        self.base.ecall(machine)
+    }
+    fn ebreak(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        self.base.ebreak(machine)
+    }
+    fn instruction_cycles(&self, inst: Instruction) -> u64 {
+        (self.cycles)(inst)
+    }
+}
+
+/// ExecutionContext composing.
+pub trait ExecutionContextExt<Mac: SupportMachine>: ExecutionContext<Mac> {
+    /// Add a syscall handler to the this context.
+    fn with_syscall<Sys>(self, syscall: Sys) -> WithSyscall<Self, Sys>
+    where
+        Self: Sized,
+        Sys: Syscalls<Mac>,
+    {
+        WithSyscall {
+            base: self,
+            syscall,
+        }
+    }
+
+    /// Replace the debugger.
+    fn with_debugger<F>(self, debugger: F) -> WithDebugger<Self, F>
+    where
+        Self: Sized,
+        // For type inference.
+        F: FnMut(&mut Mac) -> Result<(), Error>,
+    {
+        WithDebugger {
+            base: self,
+            debugger,
+        }
+    }
+
+    /// Replace the instruction cycles function.
+    fn with_cycles<F>(self, cycles: F) -> WithCyclesFunc<Self, F>
+    where
+        Self: Sized,
+        F: Fn(Instruction) -> u64,
+    {
+        WithCyclesFunc { base: self, cycles }
+    }
+
+    /// Convert the execution context to be boxed and type erased.
+    fn boxed(self) -> BoxedExecutionContext<Mac>
+    where
+        Self: Sized + Send + Sync + 'static,
+    {
+        Box::new(self)
+    }
+}
+
+impl<Mac: SupportMachine, T: ExecutionContext<Mac>> ExecutionContextExt<Mac> for T {}
--- a/tests/machine_build.rs
+++ b/tests/machine_build.rs
@@ -5,17 +5,13 @@ use ckb_vm::machine::asm::{AsmCoreMachine, AsmMachine};
 use ckb_vm::machine::{trace::TraceMachine, DefaultCoreMachine, VERSION1, VERSION2};
 use ckb_vm::registers::{A0, A7};
 use ckb_vm::{
-    DefaultMachineBuilder, Error, Register, SparseMemory, SupportMachine, Syscalls, WXorXMemory,
-    ISA_A, ISA_B, ISA_IMC, ISA_MOP,
+    DefaultMachineBuilder, Error, ExecutionContext, Register, SparseMemory, SupportMachine,
+    WXorXMemory, ISA_A, ISA_B, ISA_IMC, ISA_MOP,
 };
 
-pub struct SleepSyscall {}
+pub struct SleepContext;
 
-impl<Mac: SupportMachine> Syscalls<Mac> for SleepSyscall {
-    fn initialize(&mut self, _machine: &mut Mac) -> Result<(), Error> {
-        Ok(())
-    }
-
+impl<Mac: SupportMachine> ExecutionContext<Mac> for SleepContext {
     fn ecall(&mut self, machine: &mut Mac) -> Result<bool, Error> {
         let code = &machine.registers()[A7];
         if code.to_i32() != 1000 {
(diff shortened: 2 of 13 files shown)


src/context/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /// Replace the instruction cycles function.
+    fn with_cycles<F>(self, cycles: F) -> WithCyclesFunc<Self, F>
+    where
+        Self: Sized,
+        F: Fn(Instruction) -> u64,
+    {
+        WithCyclesFunc { base: self, cycles }
+    }
```
blckngm, line 165: I find little reason to use boxed and type erased execution contexts, but if you think it'll be useful, we can add a `boxed` method (similar to `FutureExt`) and implement `ExecutionContext` for `Box<dyn ExecutionContext>`. Maybe we should add `boxed_local` too.

mohanson:
You may consider merging this pull request; however, I must decline to do so for the following reasons:

- This pull request does not provide any benefits for CKB-VM and CKB.

- CKB-VM places significant importance on backward compatibility. Historically, upgrading CKB-VM has been a straightforward process, typically involving only the update of the version number in `Cargo.toml`. The only potentially disruptive upgrade involves modifying CKB-VM to ensure thread-safety. However, this requirement pertains to CKB itself, rather than being an inherent requirement of CKB-VM.

- Building upon the previous point, we have numerous tools and experimental codes that rely on CKB-VM. Permitting this disruptive upgrade would impose a substantial workload.

Here, I propose a compromise solution: we will create a new branch called "release-0.24-godwoken" within the CKB-VM project. This branch will be maintained in a similar manner to "release-0.24." Consequently, you can submit your commit for this pull request to the "release-0.24-godwoken" branch.

blckngm:
>     * This pull request does not provide any benefits for CKB-VM and CKB.

This makes ckb-vm machines much more flexible, supporting different lifetimes, static/dynamic dispatching, Send/Sync or not.

>     * CKB-VM places significant importance on backward compatibility. Historically, upgrading CKB-VM has been a straightforward process, typically involving only the update of the version number in `Cargo.toml`. The only potentially disruptive upgrade involves modifying CKB-VM to ensure thread-safety. However, this requirement pertains to CKB itself, rather than being an inherent requirement of CKB-VM.

This is far less disruptive than #299. Modifications needed to adopt this will be mostly mechanical. See the tests and examples.

>     * Building upon the previous point, we have numerous tools and experimental codes that rely on CKB-VM. Permitting this disruptive upgrade would impose a substantial workload.
> 
> 
> Here, I propose a compromise solution: we will create a new branch called "release-0.24-godwoken" within the CKB-VM project. This branch will be maintained in a similar manner to "release-0.24." Consequently, you can submit your commit for this pull request to the "release-0.24-godwoken" branch.

blckngm:
https://github.com/nervosnetwork/ckb/pull/3752

> There are some gray areas where some types are theoretically not Sync, such as store_transaction

CKB may actually want a `Send` but not `Sync` Machine, this change will make it possible.

blckngm:
Having a more flexible API means we can have a more stable API in the long run. We won't need to make yet another breaking change if it turns out that we need some slightly different `Send`/`Sync` bounds or lifetimes.

src/context/mod.rs
```diff
@@ -0,0 +1,152 @@
+use ckb_vm_definitions::instructions::Instruction;
+
+use super::Error;
+use crate::machine::SupportMachine;
+
+pub trait ExecutionContext<Mac: SupportMachine> {
+    fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        // We don't want to change param name to start with _ or others
```
xxuejie: Can you shed some lights on the reasons?

tests/test_resume2.rs
```diff
@@ -343,11 +343,7 @@ impl DataSource<u64> for TestSource {
 
 struct InsertDataSyscall(Arc<Mutex<Snapshot2Context<u64, TestSource>>>);
 
-impl<Mac: SupportMachine> Syscalls<Mac> for InsertDataSyscall {
-    fn initialize(&mut self, _machine: &mut Mac) -> Result<(), Error> {
-        Ok(())
-    }
-
+impl<Mac: SupportMachine> ExecutionContext<Mac> for InsertDataSyscall {
```
xxuejie, line 346: I wasn't sure on this, perhaps @mohanson's rejection partially comes from this: I could get the benefits of having `ExecutionContext` handling lifetimes, but I do believe the merits of keeping the old Syscalls trait for more complicated syscalls. Right now this PR would push for the following cases:

* Simpler syscalls must be changed to pure functions
* More complicated syscalls must implement ExecutionContext instead

What if we keep a middle ground where thread-safe syscalls can safely use Syscalls trait? One can also directly implement ExecutionContext for the referenced types, but having Syscalls in the codebase definitely aids backwards compatibility. Even better, we can keep a thread safe version of Syscalls trait and a non-thread safe version of NonThreadSafeSyscalls trait.

Simply removing Syscalls trait, IMHO, is not a good choice judging from both compatibility and code clarity reasons. Syscalls are syscalls, developers should only deal with syscalls when they needed, maybe ExecutionContext can mostly remain as a transparent concept?

xxuejie (commented):
Overall I agree this is much useful, but I think we can avoid some of the controversial parts by preserving the old Syscalls trait.

tests/test_resume2.rs
blckngm: Syscall impls don't need to be changed to pure functions. Just implement `ExecutionContext` instead of `Syscall`.

I agree that we can keep the `Syscall` trait for better compatibility, for stacking multiple `Syscall` impls like ckb.

But it will have nothing to do with thread safety. It shouldn't be necessary to declare `Syscall: Sync + Sync`. The machine types are `Send`/`Sync` when the type params are.

src/context/mod.rs
blckngm: It's for IDE completion of trait impls. I don't really recall whether IDEs remove starting underscores of args when completing trait impl methods.

It also looks better in generated docs.

xxuejie: Are there other projects that also leverage the same techniques? I wasn't sure if we should aid what the IDE should do well here.

tests/test_resume2.rs
xxuejie: Let's break it down:

* I don't think syscalls should implement `ExecutionContext`, syscalls should simply implement Syscalls trait. In fact, I want to hide `ExecutionContext` as much as possible. Users of ckb-vm should not need to know `ExecutionContext` in 99% of the cases, all they need to deal with, is methods in DefaultMachineBuilder. One example here, is I think we should see `ExecutionContext` much like the parameter `A` used to declare allocator in [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html)
* The fundamental goal here, is we want to make sure `AsmMachine` as well as `TraceMachine` are thread safe, so we can safely move them among threads in CKB. It is just due to the previous design that Syscalls must be declared as `Sync + Send` to achieve this goal. So as long as `AsmMachine` / `TraceMachine` can be `Send + Sync` with the help of `ExecutionContext`, I think we can safely remove `Sync + Send` on Syscalls trait directly, but I wasn't so familiar with lifetime gymnastics to draw a conclusion here. That's why I suggested a pair of Syscalls / NonThreadSafeSyscalls in previous comment. If you think AsmMachine can be `Send + Sync` even if Syscalls traits are not `Send + Sync`, I'm personally good with removing those type parameters on Syscalls.

src/context/mod.rs
blckngm: Not sure. But why not?

blckngm: https://internals.rust-lang.org/t/unused-parameters-in-default-trait-impls/10984

tests/test_resume2.rs
blckngm: * It's not like really `Allocator`, where the default global one is used 99% of the times. When you don't need to name to machine type, it doesn't matter, you can just use the builder methods to compose a machine. When you do, you need to decide whether you want to define a type that implements `ExecutionContext` or use `Box<dyn>` or maybe use impl trait if it's just in return type.
* It's not necessary. All the tests that move a machine to another thread are still working. And if you generate docs for this you'll see:

impl<Inner, ExecutionContext> [Send](https://doc.rust-lang.org/1.67.1/core/marker/trait.Send.html) for [DefaultMachine](file:///home/ygh/src/ckb-vm/target/doc/ckb_vm/machine/struct.DefaultMachine.html)<Inner, ExecutionContext>where
    ExecutionContext: [Send](https://doc.rust-lang.org/1.67.1/core/marker/trait.Send.html),
    Inner: [Send](https://doc.rust-lang.org/1.67.1/core/marker/trait.Send.html),
impl<Inner, ExecutionContext> [Sync](https://doc.rust-lang.org/1.67.1/core/marker/trait.Sync.html) for [DefaultMachine](file:///home/ygh/src/ckb-vm/target/doc/ckb_vm/machine/struct.DefaultMachine.html)<Inner, ExecutionContext>where
    ExecutionContext: [Sync](https://doc.rust-lang.org/1.67.1/core/marker/trait.Sync.html),
    Inner: [Sync](https://doc.rust-lang.org/1.67.1/core/marker/trait.Sync.html),

blckngm: Even when using `Box<dyn>` it shouldn't be necessary to declare `Syscalls: Send + Sync`. Instead use `Box<dyn Syscalls + Send + Sync>`.

src/context/mod.rs
blckngm: Seems that it's still useful e.g. when you need to name the context type but cannot. Adding it now.

blckngm:
There's also noticeable (at least 5%, sometimes I see 9%) performance gain for the rust interpreter:

```
interpret secp256k1_bench
                        time:   [7.0647 ms 7.0687 ms 7.0729 ms]
                        change: [-5.6207% -5.4470% -5.2768%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```

tests/test_resume2.rs
blckngm: I have restored the `Syscalls` trait and `impl Syscalls for Box<T> where T: Syscalls`. So existing code that uses only `syscall` and `instruction_cycle_func` (this is because `Box<F>` implements `Fn(...)` if `F` implements `Fn(...)`) and doesn't care about the machine type will work without modification, like the vm runner example and many tests. (But it'll be static dispatching instead of dynamic dispatching. It's also recommended to remove the `Box` if it isn't really necessary.)

blckngm:
Existing code that uses `Vec<Box<dyn Syscalls>>` can use `builder.syscall(vec_of_syscalls)`.

src/context/mod.rs
xxuejie: I feel like we are self-inventing a weird pattern for a problem that 1) does not exist, or; 2) conventionally gets resolved in a different way.

I've seen the usage of `#[allow(unused_variables)]`, or the use of `let _ = machine;` before. But combining a clippy rule with drop here is not something people used earlier. To avoid future confusion, I would recommend switch to either of the above 2 solutions, also leave a more complete reason in comments as why we don't want to change param name to start with _

tests/machine_build.rs
```diff
@@ -5,17 +5,13 @@ use ckb_vm::machine::asm::{AsmCoreMachine, AsmMachine};
...
 
-pub struct SleepSyscall {}
-
-impl<Mac: SupportMachine> Syscalls<Mac> for SleepSyscall {
-    fn initialize(&mut self, _machine: &mut Mac) -> Result<(), Error> {
-        Ok(())
-    }
+pub struct SleepContext;
 
+impl<Mac: SupportMachine> ExecutionContext<Mac> for SleepContext {
```
xxuejie, line 14: Here we are still requiring syscalls to implement ExecutionContext, which is quite some unnecesary change for the API side. I would suggest we have Syscalls implement ExecutionContext. If as you mentioned, one needs to distinguish between ExecutionContext, and `Box<dyn>` , we should let them distinguish between a `Syscalls` and `Box<dyn>` or slightly rename them if you like.

tests/test_asm.rs
```diff
@@ -71,7 +73,7 @@ pub struct CustomDebugger {
     pub value: Arc<AtomicU8>,
 }
 
-impl<Mac: SupportMachine> Debugger<Mac> for CustomDebugger {
+impl<Mac: SupportMachine> ExecutionContext<Mac> for CustomDebugger {
```
xxuejie: Similarly, if we really care about API compatibiliy, we should preserve the original Debugger trait, and have Debugger trait implement ExecutionContext.

tests/test_resume.rs
```diff
@@ -242,10 +242,22 @@ impl MachineTy {
     }
 }
 
+struct ConstantCyclesCtx;
+
+impl<Mac: SupportMachine> ExecutionContext<Mac> for ConstantCyclesCtx {
```
xxuejie, line 247: Similar to the above

xxuejie (commented):
Also, it would be much appreciated if some tests can be added to preserve the thread safey of AsmMachine and TraceMachine (with syscalls similar to how we would implement them in CKB), for reference reasons and future-proof reasons.

tests/test_resume2.rs
xxuejie: > It's not like really Allocator, where the default global one is used 99% of the times. When you don't need to name to machine type, it doesn't matter, you can just use the builder methods to compose a machine. When you do, you need to decide whether you want to define a type that implements ExecutionContext or use Box<dyn> or maybe use impl trait if it's just in return type.

We will have to agree to disagree, the above arguments, to me, is proof that ExecutionContext still is leaky abstraction that requires slightly more work. Yes in the new builder design, from now and then there might be the need to explicitly specify a type, but my question is: does one need to specify a full `ExecutionContext` type? Can we stop at syscall or debugger level?

> All the tests that move a machine to another thread are still working.

I wasn't sure on this, do we have tests like this in this repository?

src/context/mod.rs
blckngm: It's not a big issue, but definitely exists. Drop and `let _ =` is the same but drop is more future proof than `let _` because clippy already have lints for drop but don't have lints for `let _` now. If clippy adds some lint for `let _ = drop_or_ref` in the future, we would need to add a clippy allow too.

Changing to `#[allow(unused_variables)]` might work, but it'll make the signature a lot noisier.

Comment can be added.

tests/test_resume2.rs
blckngm: The alternative is to have four type parameters, which is IMO a bit too many.

Or maybe keep using `Box<dyn>` for debugger/cycles function like the previous PR. But it's actually really limiting because you are then restricted to `Send + Sync + 'static` Debuggers. IMO it's more leaky than having a single context trait.

We do have tests with threads, just grep for spawn.

tests/machine_build.rs
blckngm: This is when you need to name the context type or machine type. If you don't expand `SleepSyscall` to `SleepContext` you'd need to use `WithCyclesFunc<..., fn (Instruction) -> u64>` or `Box<dyn>` or impl trait in return type, which are all possible but not really pretty.

tests/test_resume.rs
blckngm: This is when you need to name the context type. This is personally preferable to `WithCyclesFunc<(), fn ...>` or `Box dyn`.

tests/machine_build.rs
blckngm: It's possible for T: Syscalls -> T: ExecutionContext, but a cycles func is usually also used when there's custom syscalls. It can be added nevertheless.

tests/test_asm.rs
blckngm: OK. Will do.

tests/test_resume2.rs
xxuejie: I don't really see why we need 4 type parameters.

tests/machine_build.rs
blckngm: This is actually not possible: conflicting implementations of trait `context::ExecutionContext<_>` for type `Box<(dyn context::ExecutionContext<_> + Send + Sync + 'static)>`
downstream crates may implement trait `syscalls::Syscalls<_>` for type `std::boxed::Box<(dyn context::ExecutionContext<_> + std::marker::Send + std::marker::Sync + 'static)>`

tests/test_resume2.rs
blckngm: I mean `DefaultMachine<Inner, Syscalls, Debugger, InstCyclesFunc>`.

tests/test_asm.rs
blckngm: It's similarly not possible to

```rust
impl<Mac, T> ExecutionContext<Mac> for T
where
    Mac: SupportMachine,
    T: Debugger<Mac>,
```

due to conflict implementations.

tests/test_resume.rs
blckngm: I don't really get what you want the context type to look like here. `impl ExecutionContext for F where F: Fn(Instruction) -> u64` and then `SomeMachine<fn(Instruction) -> u64>`? Doesn't look better than this IMO.

src/context/mod.rs
blckngm: Comment added.

tests/machine_build.rs
```diff
@@ -27,15 +23,17 @@ impl<Mac: SupportMachine> Syscalls<Mac> for SleepSyscall {
...
 
 #[cfg(has_asm)]
-pub fn asm_v1_imcb(path: &str) -> AsmMachine {
+pub fn asm_v1_imcb(path: &str) -> AsmMachine<SleepContext> {
     let buffer: Bytes = std::fs::read(path).unwrap().into();
     let asm_core = AsmCoreMachine::new(ISA_IMC | ISA_B, VERSION1, u64::max_value());
     let core = DefaultMachineBuilder::<Box<AsmCoreMachine>>::new(asm_core)
-        .instruction_cycle_func(Box::new(constant_cycles))
-        .syscall(Box::new(SleepSyscall {}))
+        .context(SleepContext)
```
blckngm, line 36: See all these places where syscalls and instruction cycles are set to the same tuple, IMO it's a strong indication that a unified context actually makes sense.

xxuejie:
In general I think now the PR looks better, but there are still things that puzzle me: I just realized that with this nested ExecutionContext design, one is forced into either of the following 2 scenarios:

* Use a single ExecutionContext to handle everything, TBH I think this hurts extensibility. I wonder if it's better to design a single context that wraps everything.
* Limit the number of individual syscall objects to use, otherwise the context type becomes super hard to type

Personally, I don't really buy either of the solutions here, I could get the benefit from having syscall that might contain lifetime, but the solution now here, IMHO still contains quirks I'm not sure if we want to take in. So personally, I recommend to leave this PR for now to give it some more time more thinking.

EDIT: part of me even wonder if it's better when DefaultMachine simply contains a context field of type `Box<dyn ExecutionContext>`, so we can completely erase the type parameter on top of AsmMachine or TraceMachine, leaving a cleaner type definition.

xxuejie:
Also, I've repeatedly run the benchmarks, I cannot get a reliable performance boost, so my conclusion is that this PR is more of cosmetic change, I wasn't sure if there is a real performance improvement.

blckngm:
When using multiple `Box<dyn Syscalls>` and a cycles function the cotnext type will be like `WithCyclesFunc<WithSyscalls<(), Vec<BoxedSyscalls>>, fn (Instruction) -> u64>`. It's long but you can always give it an alias or hide the whole type in a Box<dyn> or use return type impl trait.

> EDIT: part of me even wonder if it's better when DefaultMachine simply contains a context field of type Box<dyn ExecutionContext>, so we can completely erase the type parameter on top of AsmMachine or TraceMachine, leaving a cleaner type definition.

The whole point is to NOT hard code using `Box<dyn>`, because it'll force a single `Send`/`Sync`/`lifetime`. But this is allowed by the current design, see `BoxedExecutionContext` and the boxed method of `ExecutionContextExt` and the builder.

blckngm:
It's also possible to use `.syscall(A).syscall(B).syscall(C).boxed().build()`, here `A`, `B`, `C` don't need to be boxed, and the final context type is boxed. This also allows inlining all the `ecall` functions.

xxuejie:
We are really talking about different things here, I could get what you aim at here, but it still puzzles me that the AsmMachine or TraceMachine struct, has a rather complicated type definition, and the only cure is a custom ExecutionContext impl struct that encapsulates all the complicated bits. This is a direction I want to avoid.

On a completely different direction, I think maybe what you need, is a different struct than the current DefaultMachine. DefaultMachine is named DefaultMachine but it does not mean there should not be a different struct containing syscalls, instruction cycle funcs, etc, or even start with a different machine struct than AsmMachine. I think that might be a better solution to your problem.

blckngm:
Having different structs would mean code duplication, which is something that can be avoided with generics (or polymorphism in general).

xxuejie:
> Having different structs would mean code duplication, which is something that can be avoided with generics (or polymorphism in general).

I agree, but apparently generics cannot solve all the quirks we have here, the code might never be in a state where we both agree onto. In that sense I would prefer a little duplication of code. Might be better to build a custom machine struct like AsmMachine for your needs.

blckngm:
I don't see much quirks except for needing to decide about the context type sometimes. For me the flexibility is definitely worth it.

But here's an alternative design that's probably more like what you would want: we can define a `DynContext` that encapsulates a `Vec<BoxedSyscalls>` and an `Option<Box<dyn Debugger>>` and a `Box<InstructionCycleFunc>`. If we use this as the **default type parameter**, we can stay fully compatible with previous design and API. And when someone needs `!Send` or not `'static` `Syscalls`, he defines his own context. What do you think of this?

src/context/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
+        // We don't want to change the param name to start with an
+        // underscore(_). It doesn't look good in docs. Also when someone
+        // implements this method with IDE completion, they will need to remove
+        // the staring underscore.
+        #[allow(clippy::drop_ref)]
+        drop(machine);
+        Ok(())
+    }
+    /// Return true if the syscall has been processed. If a module returns
```
quake: suggest to use `allow unused_variables` in trait level:
```suggestion
#[allow(unused_variables)]
pub trait ExecutionContext<Mac: SupportMachine> {
    fn initialize(&mut self, machine: &mut Mac) -> Result<(), Error> {
        Ok(())
    }
```

blckngm: I don't think this is better because it's allowing more stuff than necessary and the intent is less clear.

This [comment](https://internals.rust-lang.org/t/unused-parameters-in-default-trait-impls/10984/7) suggests using #[allow(unused_variables)] on individual parameters, but IMO it makes the signature a lot messier.

quake: rust team member suggest to use unused_variables: https://github.com/rust-lang/rust/issues/26487#issuecomment-169828205

blckngm: It's suggesting `#[allow]`, and we are using `#[allow]` on the most specific item.

src/context/mod.rs
blckngm: After a bit of research, I learned that [`let _ =` is not the same as `drop`](https://news.ycombinator.com/item?id=22449960) and is a bit more idiomatic for ignoring stuff. So it's unlikely that clippy will add lint for these, so I've changed this to use `let _ =`.

It can also be `_ = machine` but that's a relatively new language feature and will probably look foreign to many.

There's also concern about [accidentally ignoring other stuff](https://users.rust-lang.org/t/what-is-the-best-way-to-ignore-a-result/55187) when you only want to ignore a result. It's not really relevant here but is something that should be kept in mind when ignoring results. Personally I sometimes define a `ignore_error<E>(err: Result<(), E>)` function specifically for ignoring errors.

xxuejie:
> I don't see much quirks except for needing to decide about the context type sometimes. For me the flexibility is definitely worth it.
> 
> But here's an alternative design that's probably more like what you would want: we can define a `DynContext` that encapsulates a `Vec<BoxedSyscalls>` and an `Option<Box<dyn Debugger>>` and a `Box<InstructionCycleFunc>`. If we use this as the **default type parameter**, we can stay fully compatible with previous design and API. And when someone needs `!Send` or not `'static` `Syscalls`, he defines his own context. What do you think of this?

This makes me even more firmly believe that the best solution here shall be a different machine struct. To me, this new context thing is really weird, cuz the DefaultMachine struct should already be the context. Having a default DynContext thing would definitely makes me wonder if we should just integrate it into machine struct.

blckngm:
I don't really see why we need a different struct or what the different struct would look like without generics. Fundamentally we cannot support flexible Send/Sync bounds without generics.

(Lifetime is a bit different. We can actually have flexible lifetime with `Box<dyn Trait + 'a>`, the `'a` can be specified as `'static` if desired. So thinking back it probably wasn't a good idea to remove the lifetime parameter if we are to keep using `Box<dyn>`.)

mohanson:
Judging from the current discussion, it is difficult for us to convince each other.

So why not consider my original suggestion and let's create a new branch named: `release-0.24-godwoken` (maybe it should be called `release-0.24-l2` now) @blckngm @xxuejie @quake

xxuejie:
To me this is just a matter of design preference that is determined by previous backgrounds. There is no way either of the 2 sides can persuade the other. Personally, my opinion is NOT to move forward with this PR. The generic based solution, is just not the best solution if you ask me. If reference based syscalls are really needed, my suggestion is to write a different struct of DefaultMachine, or AsmMachine that allows syscalls that do not satisfy Send + Sync, in fact this is why I named DefaultMachine DefaultMachine, it is expected there will be other machine structs from day 1. If code duplication really is a problem, common methods can then be extracted.

This will be my final comments for now.

blckngm:
To me all the previous churns removing and adding the lifetime parameter and changing the Send/Sync bounds in the `Box` definitely show that the current design is unnecessarily restrictive.

Also the performance gain for rust interpreter is consistently reproducible on my machine. It's also more profound when enabling LTO.

It's certainly possible to put it in a branch. We'll also need to rename the symbols, which I already have a patch for. Do you think we should use this branch which is based on develop or cherry pick to release-0.24?

mohanson:
I have reviewed the created branch: https://github.com/nervosnetwork/ckb-vm/tree/release-0.24-l2

If you genuinely desire to reintroduce lifetime parameters, you can submit a pull request to this branch. Once submitted, I will maintain this branch similarly to how we maintain the main branch, "release-0.24."

Lastly, I would like to reiterate my point. It is my hope that both the L2 project and CKB can utilize the same version of CKB-VM, thereby reducing the workload for both parties. 

@blckngm

blckngm:
This PR is now only minimally breaking, accommodating for existing patterns like `Vec<Box<dyn Syscalls>>`. I've also proposed the `DynContext` default parameter approach, which is fully backward compatible. I do believe you should consider incorporating one of them or something similar, not just for what godwoken needs, but also for the long term flexibility and stability for the ckb-vm API.

PR #356, opened 2023-07-11
feat!(0.24-l2): ExecutionContext trait and rename symbols

blckngm:
* Use generic type parameter instead of `Box<dyn>` to support different !Send/!Sync/not 'static Syscalls.
* Rename symbols so this can be used with crates.io version in one binary.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -6,6 +6,11 @@
 fn main() {
     use std::env;
 
+    let version = env::var("CARGO_PKG_VERSION")
+        .unwrap()
+        .replace(['.', '-'], "_");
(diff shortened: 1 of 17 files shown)


mohanson (approved):
LGTM

blckngm:
Thanks for merging this, @mohanson.

Can you make sure that this branch is protected so it won't be accidentally deleted? (asked by @Flouse)

mohanson:
> Thanks for merging this, @mohanson.
> 
> Can you make sure that this branch is protected so it won't be accidentally deleted? (asked by @Flouse)

Of course

PR #374, opened 2023-08-22
fix: Fix a bug that FastMemory does not correctly initialize memory

xxuejie:
Thanks to @mohanson for providing the failure test case

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -198,13 +198,20 @@ struct FastMemory<'a>(&'a mut Box<AsmCoreMachine>);
 
 impl<'a> FastMemory<'a> {
     fn prepare_memory(&mut self, addr: u64, size: u64) -> Result<(), Error> {
-        let aligned_start = round_page_down(addr);
-        if aligned_start < addr {
-            check_memory(self.0, aligned_start >> RISCV_PAGE_SHIFTS);
+        let frame_start = addr >> MEMORY_FRAME_SHIFTS << MEMORY_FRAME_SHIFTS;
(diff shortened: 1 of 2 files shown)


src/machine/asm/mod.rs
```diff
@@ -198,13 +198,20 @@ struct FastMemory<'a>(&'a mut Box<AsmCoreMachine>);
...
     fn prepare_memory(&mut self, addr: u64, size: u64) -> Result<(), Error> {
-        let aligned_start = round_page_down(addr);
-        if aligned_start < addr {
-            check_memory(self.0, aligned_start >> RISCV_PAGE_SHIFTS);
+        let frame_start = addr >> MEMORY_FRAME_SHIFTS << MEMORY_FRAME_SHIFTS;
+        // There is some memory space between the start of the first memory
+        // frame touched, and the starting address of memory to be written. We
+        // will need to initialize the last memory frame.
+        if frame_start < addr {
+            check_memory(self.0, addr >> RISCV_PAGE_SHIFTS);
```
XuJiandong, line 206: Is it possible to perform a direct check of the memory at 'addr' and 'addr+size'? This action would allow check_memory to handle the remaining initialization tasks.

XuJiandong (changes requested):

XuJiandong (dismissed):

src/machine/asm/mod.rs
xxuejie: The extra check here is used to rule out some load operations that are already aligned, however, I do not know the percentage of such ELFs, chances are we are fine doing either way.

mohanson (approved):

PR #393, opened 2023-12-15
chore: Add useful information to certain Error variants

xxuejie:
This commit revisits Error type with additional information that would help debugging, such as addresses to access in memory errors, and segment vaddr in Elf errors.

To preserve the required information, one additional assembly line is required in each WRITE operation of x64 implementation. The arm64 code does not suffer from any additional assembly code, in fact, we also took the liberty to reduce a few unnecessary arm64 assembly lines.


src/machine/asm/mod.rs
```diff
@@ -369,25 +379,34 @@ impl Memory for Box<AsmCoreMachine> {
...
-            return Err(Error::MemOutOfBound);
+        if round_page_up(size) != size {
+            return Err(Error::MemPageUnalignedAccess(addr + size));
         }
+
+        if addr > self.memory_size() as u64 {
+            return Err(Error::MemOutOfBound(addr, OutOfBoundKind::Memory));
+        }
+        if size > self.memory_size() as u64 || addr + size > self.memory_size() as u64 {
+            return Err(Error::MemOutOfBound(addr + size, OutOfBoundKind::Memory));
```
mohanson: If size is a large number, then addr + size will cause overflow. 

This problem can be easily reproduced using fuzz testing by `cargo +nightly fuzz run asm`.

xxuejie: This is a simple reordering of the original code, if here `addr + size` will overflow, it means the original code here also can cause overflow.

So if you were saying we have an overflowing problem, we should fix in the existing code, then reorder the code here accordingly

mohanson: Let's read the original code carefully, we will find that when the size is very large, addr + size will not be calculated; 
But in the new code, addr + size will be calculated here at `Error::MemOutOfBound(addr + size`

xxuejie: `if size > self.memory_size() as u64 || addr + size > self.memory_size() as u64`

Looking at this line, I don't believe that when size is too large, `addr + size` will be calculated, the left branch of `||` will be true and the right branch will be shortcuted.

The only case I can think of is that `addr` gets too large, causing the overflow to happen, but the question remains: on line `389` we have asserted that `addr` will be within range. I still think we might be seeing a case where the original code has issues as well.

XuJiandong: It triggers overflow here:
```
Error::MemOutOfBound(addr + size)
```
Note the `+` operator.

xxuejie: Ahhh I see, thanks

mohanson (approved):
LGTM

XuJiandong (approved):

PR #425, opened 2024-03-26
fix: Higher address bits are truncated in x64 assembly machine

xxuejie:
Note this is a bug that affects behavior, we can only fix it in VERSION2 and beyond


mohanson (approved):
LGTM

PR #451, opened 2025-01-17
test: Experiment the new native arm64 runner in CI

xxuejie:
Since arm64 runner is still quite new, we are keeping the old arm tests running on x64 machiens for a while

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -24,30 +24,50 @@ jobs:
         run: make ci-generated
 
   linux-x86-ci:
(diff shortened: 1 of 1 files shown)


xxuejie:
~~This is now pending on a permission denied error in the native arm64 runner, see [here](https://github.com/orgs/community/discussions/148648) for updates.~~

As of now, github has fixed the issue with arm64 runners

XuJiandong (approved):

mohanson (approved):

PR #461, opened 2025-03-07
perf: optimize cmp zero and jump with test

quake:
When checking if a register is zero (`cmp $0, reg`), `test reg, reg` achieves the same result while using fewer bytes in encoding.

I tested on 3 different machines, got a 1%~4% improvement.


xxuejie (approved):
Similar to previous one, please also cherry-pick to develop branch

mohanson (approved):

PR #469, opened 2025-03-21
feat: Add new DefaultMachineRunner trait for initialization work

xxuejie:
Previously, all different machines in ckb-vm might use slightly different initialization process. This commit introduces a new DefaultMachineRunner trait, so the same initialization & usage workflow can be used for all types implementing this trait.


mohanson (approved):
LGTM

quake (approved):

PR #470, opened 2025-03-21
feat: Add new DefaultMachineRunner trait for initialization work

xxuejie:
Previously, all different machines in ckb-vm might use slightly different initialization process. This commit introduces a new DefaultMachineRunner trait, so the same initialization & usage workflow can be used for all types implementing this trait.

This is a cherry-pick of https://github.com/nervosnetwork/ckb-vm/pull/469 to the develop branch. 2 more changes are applied on the develop branch:

* Since AsmCoreMachine is now constant, we can simply use AsmCoreMachine instead of `Box<AsmCoreMachine>`.
* https://github.com/nervosnetwork/ckb-vm/pull/343 was mostly reverted. Upon more thinking, I think the change earlier does not make sense now. It's better we provide unified initialization for the whole machine, not just for the memory part.

Due to the new InstDecoder / TraceDecoder structure introduced in develop branch, more changes are required in the develop branch than 0.24.x releases. Some of those changes include:

* Assembly machines now require TraceDecoder instead of InstDecoder. This means we cannot use DefaultMachineBuilder to build DefaultMachine structures for both types of VMs. The assembly machine types require more constraints. Given this consideration, DefaultMachineBuilder is removed in this commit. 2 new types RustDefaultMachineBuilder and AsmDefaultMachineBuilder are added for 2 types of VMs respectively.
* A new set of types that begin with `Abstract`, such as AbstractDefaultMachineBuilder, AbstractAsmMachine, AbstractTraceMachine are introduced. One is only expected to use those types when one wants to tweak the inst/trace decoder used. Otherwise, one is safe to pick AsmMachine/AsmDefaultMachineBuilder or TraceMachine/RustDefaultMachineBuilder.

So the current workflow is like following:

1. Pick a type implementing DefaultMachineRunner traits. Most likely, one would pick either AsmMachine or TraceMachine.
2. Instantiate a core machine using DefaultMachineRunner::Inner::new.
3. Based on the type picked in step 1, use AsmDefaultMachineBuilder or RustDefaultMachineBuilder to build a default machine with possible instruction cycle func, syscalls and debugger plugin.
4. Finally, use DefaultMachineRunner::new to create the final runnable machine from the default machine.

For advanced usages where the decoder is customized, one can use AbstractAsmMachine or AbstractTraceMachine type. Then use AbstractDefaultMachineBuilder to build the default machine. And finally create the runnable machine. All those types accept an additional InstDecoder / TraceDecoder trait impl that allows one to customize the decoder for optimizations.

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -9,29 +9,32 @@ use ckb_vm_definitions::{
         RET_EBREAK, RET_ECALL, RET_INVALID_PERMISSION, RET_MAX_CYCLES_EXCEEDED, RET_OUT_OF_BOUND,
         RET_PAUSE, RET_SLOWPATH,
     },
-    ISA_MOP, MEMORY_FRAME_PAGE_SHIFTS, RISCV_GENERAL_REGISTER_NUMBER, RISCV_PAGE_SHIFTS,
+    ISA_MOP, MEMORY_FRAMESIZE, MEMORY_FRAME_PAGE_SHIFTS, RISCV_GENERAL_REGISTER_NUMBER,
+    RISCV_PAGE_SHIFTS,
 };
 use rand::{prelude::RngCore, SeedableRng};
+use std::alloc::{alloc, alloc_zeroed, Layout};
+use std::marker::PhantomData;
+use std::mem::MaybeUninit;
 use std::os::raw::c_uchar;
 
 use crate::{
-    decoder::{build_decoder, InstDecoder},
     elf::ProgramMetadata,
     error::OutOfBoundKind,
     instructions::execute_instruction,
     machine::{
         asm::traces::{decode_fixed_trace, SimpleFixedTraceDecoder, TraceDecoder},
-        VERSION0,
+        AbstractDefaultMachineBuilder, VERSION0,
     },
     memory::{
         check_no_overflow, fill_page_data, get_page_indices, memset, round_page_down,
         round_page_up, FLAG_DIRTY, FLAG_EXECUTABLE, FLAG_FREEZED, FLAG_WRITABLE, FLAG_WXORX_BIT,
     },
-    CoreMachine, DefaultMachine, Error, Machine, Memory, SupportMachine, MEMORY_FRAME_SHIFTS,
-    RISCV_PAGESIZE,
+    CoreMachine, DefaultMachine, DefaultMachineRunner, Error, Machine, Memory, SupportMachine,
+    MEMORY_FRAME_SHIFTS, RISCV_PAGESIZE,
 };
 
-impl CoreMachine for Box<AsmCoreMachine> {
+impl CoreMachine for AsmCoreMachine {
     type REG = u64;
     type MEM = Self;
 
@@ -119,7 +122,7 @@ fn check_permission<M: Memory>(memory: &mut M, page: u64, flag: u8) -> Result<()
 
 // check whether a memory address is writable or not and mark it as dirty, `size` should be 1, 2, 4 or 8
 fn check_memory_writable(
-    machine: &mut Box<AsmCoreMachine>,
+    machine: &mut AsmCoreMachine,
     addr: u64,
     size: usize,
 ) -> Result<(), Error> {
@@ -152,7 +155,7 @@ fn check_memory_writable(
 
 // check whether a memory address is executable, `size` should be 2 or 4
 fn check_memory_executable(
-    machine: &mut Box<AsmCoreMachine>,
+    machine: &mut AsmCoreMachine,
     addr: u64,
     size: usize,
 ) -> Result<(), Error> {
@@ -183,11 +186,7 @@ fn check_memory_executable(
 }
 
 // check whether a memory address is initialized, `size` should be 1, 2, 4 or 8
-fn check_memory_inited(
-    machine: &mut Box<AsmCoreMachine>,
-    addr: u64,
-    size: usize,
-) -> Result<(), Error> {
+fn check_memory_inited(machine: &mut AsmCoreMachine, addr: u64, size: usize) -> Result<(), Error> {
     debug_assert!(size == 1 || size == 2 || size == 4 || size == 8);
     let page = addr >> RISCV_PAGE_SHIFTS;
     if page as usize >= machine.memory_pages() {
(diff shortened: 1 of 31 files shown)


mohanson (dismissed):

src/machine/trace.rs
```diff
@@ -46,14 +47,17 @@ fn calculate_slot(addr: u64) -> usize {
...
+pub type TraceMachine<Inner> = AbstractTraceMachine<Inner, DefaultDecoder>;
 
-    factory: ThreadFactory<DefaultMachine<Inner>>,
-    traces: Vec<Trace<DefaultMachine<Inner>>>,
+pub struct AbstractTraceMachine<Inner: SupportMachine, Decoder> {
+    pub machine: DefaultMachine<Inner, Decoder>,
+
+    factory: ThreadFactory<DefaultMachine<Inner, Decoder>>,
+    traces: Vec<Trace<DefaultMachine<Inner, Decoder>>>,
+    phantom: PhantomData<Decoder>,
```
quake: I think this field is unnecessary since we have the generic type bound in other fields already.

src/machine/asm/mod.rs
```diff
@@ -653,44 +692,38 @@ extern "C" {
...
-        Self { machine }
-    }
+pub type AsmMachine = AbstractAsmMachine<SimpleFixedTraceDecoder>;
 
-    pub fn set_max_cycles(&mut self, cycles: u64) {
-        self.machine.inner.max_cycles = cycles;
-    }
+pub struct AbstractAsmMachine<Decoder: TraceDecoder> {
+    pub machine: DefaultMachine<AsmCoreMachine, Decoder>,
+    phantom: PhantomData<Decoder>,
```
quake: unnecessary field

quake (commented):
I think introducing another `TraceMachineRuner` to extend `DefaultMachineRunner` trait would be a lot simpler than the current associated type design:

```rust
trait DefaultMachineRunner {
    fn run_with_decoder<D: InstDecoder>(&mut self, decoder: &mut D);
}

trait TraceMachineRunner: DefaultMachineRunner {
    fn run_with_decoder<D: TraceDecoder>(&mut self, decoder: &mut D);
}

trait InstDecoder {}
trait TraceDecoder: InstDecoder {}
```

This allows us to remove the `Decoder` generic type bound from most structs, for example, `AbstractTraceMachine`:
```rust
pub struct AbstractTraceMachine<Inner: SupportMachine> {
    pub machine: DefaultMachine<Inner>,
    factory: ThreadFactory<DefaultMachine<Inner>>,
    traces: Vec<Trace<DefaultMachine<Inner>>>,
}
```

xxuejie:
> I think introducing another TraceMachineRuner to extend DefaultMachineRunner trait would be a lot simpler than the current associated type design

This would actually require specifying types when calling the method, see [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d346a3dbe3ca7f41307aac1a819c5106). In fact a lot of the work done in this PR, is done to make sure we don't need those extra work for specifying types. I do want to design the API that when you invoke `run` on AsmMachine, it will run on a `TraceDecoder` automatically. And it really does not make sense to call `AsmMachine` with a `InstDecoder`.

quake:
I see...
LGTM now

xxuejie:
Unnecessary PhantomData fields have been removed.

quake (approved):

xxuejie:
Just to provide a little bit more on the background of this PR, in case anyone comes back from the future: 

Some of the initial design goals are as follows:

* `DefaultMachineRunner` should the be top trait everyone should interact with. When using CKB-VM, one only needs to pick an impl of `DefaultMachineRunner` and start from there, the choices for other types should be derived, not chosen.
* Both the associated type design we use or, or trait generics(i.e., `DefaultMachineRunner<Inner: SupportMachine, Decoder: InstDecoder>`) should work for us, we chose associated type design, since one can write `<xxx as DefaultMachineRunner>::Inner` to fetch the inner type, while in the other case, one should know what the inner concrete type actually is.
* Instruction decoding is one of the majority bottleneck in current CKB-VM. A lower-hanging fruit at CKB side, is that we can cache decoded instructions across multiple runs of the same CKB script. As a result, `InstDecoder` is abstracted, which becomes a type parameter of `DefaultMachineRunner`.
* The assembly VM is designed with `traces` in mind. It must function together with a `TraceDecoder`. AsmMachine do not work together with a simple `InstDecoder`.
    + What we really need, is C++'s template specialization, where an AsmMachine could use a specialization of the `InstDecoder` trait. Unfortunately, this is not yet in stable Rust as of now(Mar 2025), maybe we will have trait specializations in Rust one day, and we can revisit the types again.
* We should have the user specify as few types as possible.

Really, the above are the original design goals.

The introduction of `Abstract***` types, and the additional `AsmDefaultMachineBuilder`, is really the result of having users specify as few types as possible, while navigating with the above design goals.

When I was using the original `DefaultMachineBuilder`, I kept hitting one of the following issues:

* The same `DefaultMachineBuilder` cannot work on both the `TraceMachine` and the `AsmMachine`, the extra trait restriction of `AsmMachine` forbades it.
* A user have to type a huge amount of characters to include the full types of `DefaultMachineBuilder`, even in the default case. I do mean it: the full type can be really long!

That's really the birth of `AbstractDefaultMachineBuilder`, `RustDefaultMachineBuilder` and `AsmDefaultMachineBuilder`.

While I feel this PR is long enough, a potential improvement could be added in the future: as of right now, one still needs initialization for 3 types:

* A chosen type implementing `DefaultMachineRunner` trait
* One of the 3 builder types
* `DefaultMachineRunner::Inner` type

Even though when someone chooses the type implementing `DefaultMachineRunner` trait, the other 2 will mostly be self-evident. One still needs to manually inistantiate all 3 types and know about them, one such initialization process can be found [here](https://github.com/nervosnetwork/ckb/blob/d2b009a7173f10447ba834481988313cb85cfe2c/script/src/scheduler.rs#L937-L956), and we don't expect other usages of the same API to differ from the provided example too much. Maybe in the future, we should add a builder API on `DefaultMachineRunner` trait itself, we hide the other 2 types behind the builder pattern. Maybe that will bring a more pleasant API to work with.

PR #482, opened 2025-06-04
refactor: Deduplicate and simplify several trace-end handling assembl…

xxuejie:
…y pieces


src/machine/asm/execute_x64.S
```diff
@@ -467,49 +468,14 @@ ckb_vm_x64_execute:
...
-  jc .exit_cycles_overflow
-  cmp CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_MAX_CYCLES(MACHINE), %rax
-  ja .exit_max_cycles_exceeded
-  movq %rax, CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_CYCLES(MACHINE)
-  addq %rdx, PC_ADDRESS
-  lea CKB_VM_ASM_TRACE_OFFSET_THREADS(TRACE), INST_PC
-  mov INST_PC, INST_ARGS
-  add $8, INST_ARGS
-  NEXT_INST
+  jmp .trace_loaded
```
mohanson, line 471: I found that the old code will never jump to the .exit_trace branch, but after the modification, it may jump to the .exit_trace branch through .trace_loaded. 

<del> (I can't confirm whether this jump is possible now). </del>

I checked the code carefully and I think it is OK.

mohanson (approved):

src/machine/asm/execute_x64.S
xxuejie: So when `TRACE_JUMP` is generated, the length of trace will always be positive, in this sense it will never taken the jump at line 448.

Ideally for `TRACE_JUMP` we don't need tests on line 447-448 but they are just trivial to run, won't cause any noticeable difference.

PR #486, opened 2025-09-11
feat: Add Nix flake for RISC-V toolchain setup

bubblepipe:
This PR adds a Nix flake script to provide an easy way to install the RISC-V toolchain required for compiling C programs to run on CKB VM. 

By running:
```
$ nix develop 
```
We can have a working `riscv64-none-elf-gcc` in our environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ Cargo.lock
 /examples/is13
 *.DS_Store
 tests/programs/deps
+flake.lock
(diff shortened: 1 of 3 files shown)


XuJiandong:
We use native clang to compile on-chain script. See https://github.com/cryptape/ckb-script-templates/blob/main/workspace/scripts/find_clang

PR #492, opened 2025-11-10
[WIP] Implement CFI

mohanson:
More information: [Design of CFI in CKB-VM](https://github.com/nervosnetwork/ckb-vm/blob/cfi/docs/CFI.md).


XuJiandong:
See dedicated fuzzing tests here: https://github.com/nervosnetwork/ckb-vm-fuzzing-test/pull/15

XuJiandong:
In LLVM 21 and 22, the following command line can be used to enable CFI support:

```
--target=riscv64-unknown-elf -march=rv64imc_zba_zbb_zbc_zbs_zicfiss1p0_zicfilp1p0 \
-menable-experimental-extensions -fcf-protection=full -mcf-branch-label-scheme=func-sig
```

There is no change in CFI support between version 21 and 22.

| Flag / Extension | Status in LLVM 22 | Notes |
| --- | --- | --- |
| `zicfilp1p0` | Still experimental | |
| `zicfiss1p0` | Still experimental | |
| `-menable-experimental-extensions` | Still required | Both extensions remain experimental, so this flag is mandatory |
| `-fcf-protection=full` | Still supported | Values: `return`, `branch`, `full`, `none` |
| `-mcf-branch-label-scheme=func-sig` | Still supported | Values: `unlabeled`, `func-sig` (default for RISC-V is now `func-sig`) |

One minor note: `func-sig` is now the default branch label scheme for RISC-V, but it is not truly implemented in LLVM 22. The flag is accepted by the Clang frontend, however the backend does not actually generate per-function-signature landing pad labels. Instead, the backend uses a single fixed label (default 0) for all functions.

Suggest not merging this PR now because:

1. CFI support for RISC-V is still experimental.
2. No improvement or new feature is introduced in LLVM 22.

We can revisit this when LLVM 23 is out.

mohanson:
In LLVM 23.1.0-rc3, use the following command to compile the program with CFI:

```text
--target=riscv64-unknown-elf \
-march=rv64imc_zba_zbb_zbc_zbs_zicfiss1p0_zicfilp1p0 \
-fcf-protection=full \
-mcf-branch-label-scheme=unlabeled
```

The only difference from LLVM 22 is `-mcf-branch-label-scheme=func-sig` is strictly prohibited.

PR #493, opened 2025-11-10
feat: impl arm64 asm runner

FrontMage:
Hi, I have been trying to run ckb-vm on my Mac. Turns out that it does not support the M1 chip in asm mode, so I tried to implement asm mode on arm64 hosts. Since arm64 hosts are becoming more and more popular, this feature might be helpful to have.
I did some basic tests, and they work, but I have noticed that this repo has very robust code, so it will be very helpful if you can review some of the code and point me in the direction of more robust code.

The team can give any feedback, even “this is garbage, start over”, is gold to me, or just tell me if I'm on the right track.

Thank you for your attention!

### DIFF
--- a/ASM_EXECUTION_FLOW.md
+++ b/ASM_EXECUTION_FLOW.md
@@ -1,0 +1,421 @@
+# ASM Code Execution Flow in CKB-VM
+
+This document explains the step-by-step process of executing RISC-V code using the Assembly (ASM) optimized execution mode in CKB-VM.
+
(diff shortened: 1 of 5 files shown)


XuJiandong:
already included: https://github.com/nervosnetwork/ckb-vm/blob/develop/src/machine/asm/execute_aarch64.S

FrontMage:
Wow, so I only need to make a dev build then? Thx!

Issue #222, opened 2021-12-10
Phase plan of the V instruction set


mohanson:
# Phase plan of the V instruction set

The V instruction set has a large number of instructions, and it is difficult for us to complete these instructions at once, so these instructions are divided into 3 parts.

- The first part needs to be implemented first
- The second part depends on the situation
- And the third part is instructions that we don’t need in the short term ( Floating / atomic instructions).

Note:
The first column: implemented or not
The second column: tested or not

# Part 1

```
[x] [ ] vsetivli     31=1 30=1 zimm10    zimm 14..12=0x7 rd 6..0=0x57
[x] [ ] vsetvli      31=0 zimm11          rs1 14..12=0x7 rd 6..0=0x57
[x] [ ] vsetvl       31=1 30..25=0x0 rs2  rs1 14..12=0x7 rd 6..0=0x57

[x] [ ] vlm.v          31..28=0 27..26=0 25=1 24..20=0xb rs1 14..12=0x0  vd 6..0=0x07
[x] [ ] vsm.v          31..28=0 27..26=0 25=1 24..20=0xb rs1 14..12=0x0 vs3 6..0=0x27

[x] [ ] vle8.v         nf 28=0 27..26=0 vm 24..20=0 rs1 14..12=0x0  vd 6..0=0x07
[x] [ ] vle16.v        nf 28=0 27..26=0 vm 24..20=0 rs1 14..12=0x5  vd 6..0=0x07
[x] [ ] vle32.v        nf 28=0 27..26=0 vm 24..20=0 rs1 14..12=0x6  vd 6..0=0x07
[x] [ ] vle64.v        nf 28=0 27..26=0 vm 24..20=0 rs1 14..12=0x7  vd 6..0=0x07
[x] [ ] vle128.v       nf 28=1 27..26=0 vm 24..20=0 rs1 14..12=0x0  vd 6..0=0x07
[x] [ ] vle256.v       nf 28=1 27..26=0 vm 24..20=0 rs1 14..12=0x5  vd 6..0=0x07
[x] [ ] vle512.v       nf 28=1 27..26=0 vm 24..20=0 rs1 14..12=0x6  vd 6..0=0x07
[x] [ ] vle1024.v      nf 28=1 27..26=0 vm 24..20=0 rs1 14..12=0x7  vd 6..0=0x07
[...]

PR #494, opened 2025-11-11
[WIP] feat: impl vadd on arm hosts

FrontMage:
Hi, this is a simple implementation of `vsetvli` and `vadd` on arm64 hosts using Neon.

These are the key changes that this PR made:
1. **Vector ISA plumbing**
   - Added `ISA_V` along with vector register constants in `definitions/src/lib.rs`.
   - Extended `AsmCoreMachine` (definitions) plus every `CoreMachine` implementation to allocate 32 vector registers with two 64-bit lanes, track `vl/vtype`, and surface the state via new trait methods (`src/machine/mod.rs`, `src/machine/asm/mod.rs`, `src/machine/trace.rs`).
   - Re-exported the new execute helpers so downstream crates/tests can reach them (`src/lib.rs`, `src/instructions/mod.rs`).

2. **Decoder and instruction support**
   - Added a dedicated decoder (`src/instructions/v.rs`) and wired it into `DefaultDecoder` whenever `ISA_V` is set (`src/decoder.rs`).
   - Defined opcodes `OP_VSETVLI`/`OP_VADD_VV` in `definitions/src/instructions.rs`, updated tagged conversions, and implemented the handlers in `src/instructions/execute.rs`.

3. **Assembly backends**
   - Regenerated `cdefinitions_generated.h` to expose the new struct offsets/ISA constants.
   - Implemented `vsetvli` and `vadd.vv` directly in the arm64 interpreter using NEON (`src/machine/asm/execute_aarch64.S`). The x64 backend now has stub labels that fall back to the slow path (`src/machine/asm/execute_x64.S`).

4. **Real binary validation**
   - Authored a hand-crafted ELF binary at `tests/programs/vector_vadd` containing `addi`, `vsetvli`, `vadd.vv`, and the usual `ecall`.
   - Added a new `vector_binary_runs_on_asm` test (plus a decoder sanity check) in `tests/test_vector.rs`. The test seeds vector registers directly, runs the asm backend on the real binary, and asserts that the produced vector lanes contain the expected sums.
---
This is still WIP and should NOT be merged.
Plz give any feedback, even “this is garbage, start over”, is gold to me, or just tell me if I'm on the right track.
Thank you for your attention, and have a nice day!


mohanson:
I know this is an instruction in the RISCV vector instruction set, but what is the purpose of this PR? We do not plan to support vector computation in CKB-VM.

FrontMage:
> I know this is an instruction in the RISCV vector instruction set, but what is the purpose of this PR? We do not plan to support vector computation in CKB-VM.

Thanks for the reply!
Sorry for the confusion, but wouldn't it be nice to have more instructions supported? 
From this issue #222 I can see there is a plan for the v instruction set.
Also, you were talking about plans. Is there a roadmap for this I can refer to?

XuJiandong:
We actually started developing RVV instructions 3 years ago, see https://github.com/nervosnetwork/ckb-vm/tree/rvv
However, this instruction set is too complex, with nearly 300 instructions in total (not including floating point instructions).
Compared to our current small instruction set implementation, we found it too bloated and complicated, so we decided to drop this instruction set.

PR #500, opened 2025-12-23
examples/check_real_memory: use a hand written buddy allocator instead of jemalloc

mohanson:
Using a hand-written allocator can provide accurate information about the memory used by a ckb-vm machine instance, which is better than using jemalloc.

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       # https://github.com/EmbarkStudios/cargo-deny-action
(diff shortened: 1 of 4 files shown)


XuJiandong:
What issue do we face? Can we use other allocator instead of `jemalloc`?

mohanson:
This library was initially discovered because it caused a cargo-deny warning. This library is no longer maintained on crates.io: https://crates.io/crates/jemallocator. Additionally, This library cannot run on Windows.

Furthermore, adding a complex dependency library just for an example is not worthwhile.

XuJiandong (approved):

PR #502, opened 2026-01-29
Allows ckb-vm to execute in assembly mode within the rv64im environment

mohanson:
This PR adds an ASM execution mode to ckb-vm on the riscv64 target. 

Any added or modified code should not introduce any side effects to existing logic, in other words, **Additive changes only; existing behavior is unchanged.**

Finally, I rewrote CI. In the past, ckb-vm would build test cases within CI; given that building test cases on RISCv64 is quite cumbersome, now all test cases in the ckb-vm-contirb package are packaged into an aritifact and directly included in the repository.


src/machine/asm/execute_riscv64.S
```diff
@@ -0,0 +1,1908 @@
...
+  LOAD_REGISTER_ADDRESS(RS2, t3)
+  ld     t2, CKB_VM_ASM_ASM_CORE_MACHINE_OFFSET_REGISTERS(t3)
+  add    t3, t0, t1
+  sltu   t4, t3, t0
+  add    t0, t3, t2
+  sltu   t1, t0, t3
+  or     t2, t4, t1
+  WRITE_RD_V2(t0, t3)
+  WRITE_RS1(t2, t3)
+  WRITE_RS2(t1, t3)
```
XuJiandong: AArch64 ADC writes in order: RS1, RS2, then RD (last).
RISC-V ADC writes in order: RD, RS1, then RS2 (last).
If register indices overlap (e.g., RD == RS1), the last write wins:
AArch64: register[RD] ends up with the result (RD is written last).
RISC-V: register[RD=RS1] ends up with carry1 | carry2 (RS1 is written after RD).
Same issue exists for SBB (AArch64: RS2, RS3, RS1, RD vs RISC-V: RD, RS1, RS2, RS3).

mohanson: During the decoder stage, we avoid having rd, rs1, and rs2 equal.

I don't think this will cause a bug, but we should still maintain a consistent write order.

src/machine/asm/execute_riscv64.S
```diff
@@ -0,0 +1,1908 @@
...
+  sd     s0, 8(sp)
+  sd     s1, 16(sp)
+  sd     s2, 24(sp)
+  sd     s3, 32(sp)
+  sd     s4, 40(sp)
+  sd     s5, 48(sp)
+  sd     s6, 56(sp)
+  sd     s7, 64(sp)
+  sd     s8, 72(sp)
+  sd     s9, 80(sp)
```
XuJiandong: The RISC-V ABI requires sp to be 16-byte aligned at function call boundaries. The prologue allocates 88 bytes (11 registers x 8), and PREPCALL allocates another 64 bytes, totaling 152 bytes. Since 152 mod 16 = 8, sp is NOT 16-byte aligned when inited_memory is called.

src/machine/asm/execute_riscv64.S
```diff
@@ -0,0 +1,1908 @@
...
+  ld     t2, CKB_VM_ASM_INVOKE_DATA_OFFSET_FIXED_TRACE_MASK(INVOKE_DATA)
+  and    t1, t1, t2
+  li     t2, CKB_VM_ASM_FIXED_TRACE_STRUCT_SIZE
+  mul    t1, t1, t2
+  ld     TRACE, CKB_VM_ASM_INVOKE_DATA_OFFSET_FIXED_TRACES(INVOKE_DATA)
+  add    TRACE, TRACE, t1
+  ld     t2, CKB_VM_ASM_TRACE_OFFSET_ADDRESS(TRACE)
+  bne    t2, t0, .exit_trace
+.trace_loaded:
+  lw     t3, CKB_VM_ASM_TRACE_OFFSET_LENGTH(TRACE)
```
XuJiandong: should use lwu.

src/machine/asm/execute_riscv64.S
mohanson: Confirmed as a bug.

src/machine/asm/execute_riscv64.S
mohanson: Yes.

XuJiandong (approved):

PR #509, opened 2026-04-03
refactor: replace byteorder crate usage with native rust methods

mohanson:
ckb-vm relied on the `byteorder` library to handle byte order. However, Rust's standard library (since version 1.44) already provides sufficient functionality to handle byte order, so we can consider removing the byteorder dependency.

https://doc.rust-lang.org/std/primitive.u32.html#method.from_le_bytes

Meanwhile, the byteorder repository appears to be no longer maintained, furthermore, the author no longer recommends its use: https://github.com/BurntSushi/byteorder?tab=readme-ov-file#alternatives

I believe we didn't choose byteorder over the Rust standard library, but simply because the ckb-vm repository was created before Rust 1.44

### DIFF
--- a/src/machine/asm/mod.rs
+++ b/src/machine/asm/mod.rs
@@ -1,6 +1,5 @@
 pub mod traces;
 
-use byteorder::{ByteOrder, LittleEndian};
 use bytes::Bytes;
 pub use ckb_vm_definitions::asm::AsmCoreMachine;
 use ckb_vm_definitions::{
@@ -571,13 +570,17 @@ where
     fn execute_load16(&mut self, addr: u64) -> Result<u16, Error> {
         check_memory_executable(self, addr, 2)?;
         let slice = cast_ptr_to_slice(self, self.as_ref().memory_ptr, addr as usize, 2);
-        Ok(LittleEndian::read_u16(slice))
+        let mut bytes = [0u8; 2];
+        bytes.copy_from_slice(slice);
+        Ok(u16::from_le_bytes(bytes))
     }
 
     fn execute_load32(&mut self, addr: u64) -> Result<u32, Error> {
         check_memory_executable(self, addr, 4)?;
         let slice = cast_ptr_to_slice(self, self.as_ref().memory_ptr, addr as usize, 4);
-        Ok(LittleEndian::read_u32(slice))
+        let mut bytes = [0u8; 4];
+        bytes.copy_from_slice(slice);
+        Ok(u32::from_le_bytes(bytes))
     }
 
     fn load8(&mut self, addr: &u64) -> Result<u64, Error> {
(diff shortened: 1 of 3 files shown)


mohanson:
The CI failure is not an issue with this PR.

src/machine/asm/mod.rs
```diff
@@ -571,13 +570,17 @@ where
     fn execute_load16(&mut self, addr: u64) -> Result<u16, Error> {
         check_memory_executable(self, addr, 2)?;
         let slice = cast_ptr_to_slice(self, self.as_ref().memory_ptr, addr as usize, 2);
-        Ok(LittleEndian::read_u16(slice))
+        let mut bytes = [0u8; 2];
+        bytes.copy_from_slice(slice);
```
XuJiandong, line 574: I don't know if there is any performance issue. It might be better to use like below:
```rust
u16::from_le_bytes(slice[..2].try_into()?);
```
Suggest testing both methods. If there is no performance issue, I think both are OK.

mohanson: There is no difference between them, they are usually translated into completely equivalent machine code under release.

```
parse_copy_from_slice   time:   [1.4253 ns 1.4295 ns 1.4338 ns]

parse_try_into          time:   [1.4202 ns 1.4240 ns 1.4293 ns]
```

```
Disassembly of section .text._ZN2rs14bench_try_into17he74585ab25055825E:

0000000000000000 <_ZN2rs14bench_try_into17he74585ab25055825E>:
   0:   48 83 fe 01             cmp    $0x1,%rsi
   4:   76 04                   jbe    a <_ZN2rs14bench_try_into17he74585ab25055825E+0xa>
   6:   0f b7 07                movzwl (%rdi),%eax
   9:   c3                      ret
   a:   50                      push   %rax
   b:   48 89 f2                mov    %rsi,%rdx
   e:   48 8d 0d 00 00 00 00    lea    0x0(%rip),%rcx        # 15 <_ZN2rs14bench_try_into17he74585ab25055825E+0x15>
  15:   be 02 00 00 00          mov    $0x2,%esi
  1a:   31 ff                   xor    %edi,%edi
  1c:   ff 15 00 00 00 00       call   *0x0(%rip)        # 22 <_ZN2rs14bench_try_into17he74585ab25055825E+0x22>

Disassembly of section .text._ZN2rs21bench_copy_from_slice17h8d9e7a3cf9972a64E:

0000000000000000 <_ZN2rs21bench_copy_from_slice17h8d9e7a3cf9972a64E>:
   0:   48 83 fe 01             cmp    $0x1,%rsi
   4:   76 04                   jbe    a <_ZN2rs21bench_copy_from_slice17h8d9e7a3cf9972a64E+0xa>
   6:   0f b7 07                movzwl (%rdi),%eax
   9:   c3                      ret
   a:   50                      push   %rax
   b:   48 89 f2                mov    %rsi,%rdx
   e:   48 8d 0d 00 00 00 00    lea    0x0(%rip),%rcx        # 15 <_ZN2rs21bench_copy_from_slice17h8d9e7a3cf9972a64E+0x15>
  15:   be 02 00 00 00          mov    $0x2,%esi
  1a:   31 ff                   xor    %edi,%edi
  1c:   ff 15 00 00 00 00       call   *0x0(%rip)        # 22 <_ZN2rs21bench_copy_from_slice17h8d9e7a3cf9972a64E+0x22>
```

quake: considering it's a hot path, we may try using unsafe:

```rust
pub fn load2(mem: &[u8], addr: usize) -> u16 {
    unsafe {
        let ptr = mem.as_ptr().add(addr) as *const u16;
        std::ptr::read_unaligned(ptr)
    }
}
```

which compiles to
```asm
load2:
        movzx   eax, word ptr [rdi + rdx]
        ret
```

compares the safe version adds bounds checks and branches, while the unsafe version is just the raw load.
```rust
pub fn load1(mem: &[u8], addr: usize) -> u16 {
    u16::from_le_bytes([mem[addr], mem[addr + 1]])
}
```

```asm
load1:
        push    rax
        cmp     rdx, rsi
        jae     .LBB0_3
        lea     rax, [rdx + 1]
        cmp     rax, rsi
        jae     .LBB0_2
        movzx   eax, word ptr [rdi + rdx]
        pop     rcx
        ret
```

mohanson: @quake We cannot assume that the host machine is little-endian, as this approach will cause problems with big-endian host machines.

quake: Big-endian machines are extremely rare and rust supports cfg(target_endian) for compile time branching

XuJiandong (approved):

PR #510, opened 2026-04-07
Update README.md to include support for "A" extension

Matt-RUN-CKB:
CKB2023 added support for "A" atomic instructions in CKB VM v2, AI agents are not picking this up, likely due to it being missed in the README

Updated to "IMACB"

https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0051-ckb2023/0051-ckb2023.md#ckb-vm-v2

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 ## About CKB VM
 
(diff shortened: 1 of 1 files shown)


XuJiandong:
This RFC is outdated and should be updated. A extension is removed before that hardfork.

Issue #511, opened 2026-04-09
ckb-vm's build.rs on windows-msvc using gcc? how about clang-cl?


eval-exec:
https://github.com/nervosnetwork/ckb-vm/blob/c4ae777214605b792ea74026b6abdeddb77c7310/build.rs#L40-L50

how about using `clang-cl` to replace `gcc` here?
https://clang.llvm.org/docs/MSVCCompatibility.html

mohanson:
I think we should do it this way, but it's unclear whether we can use `clang` alone (clang-cl may not be necessary). I think we can, but it needs testing.

PR #512, opened 2026-04-13
Rewrite build.rs

mohanson:
1. Build using clang instead of gcc on Windows.
2. The code in build.rs has been significantly simplified. 
3. The test worked fine on SP1.

Ref https://github.com/nervosnetwork/ckb-vm/issues/511

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -7,6 +7,7 @@ use std::env;
 
 fn main() {
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_default();
+    let target_family = env::var("CARGO_CFG_TARGET_FAMILY").unwrap_or_default();
     let support_asm = ["aarch64", "riscv64", "x86_64"].contains(&target_arch.as_str());
     if !support_asm {
         if cfg!(feature = "asm") {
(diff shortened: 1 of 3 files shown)


eval-exec (approved):

mohanson:
@eval-exec Do you want this change to be included in the release version?

cc @chenyukang

build.rs
```diff
@@ -3,61 +3,50 @@
...
+                .file("src/machine/asm/execute_aarch64.S")
+                .compile("asm");
         }
-
-        build.include("src/machine/asm").compile("asm");
-
-        println!("cargo:rustc-cfg=has_asm")
+        "riscv64" => {
+            cc::Build::new()
+                .compiler(clang_finder::find())
```
eval-exec: Not sure do we really need to hardcode `clang` and ignore `CC` env var here?
`cc-rs` allow cargo usage like: `CC=clang cargo build .....`

mohanson: Good idea, perhaps we don't need to specify compiler.

mohanson: However, if we do this, we will always need to set the CC environment variable when compiling on Windows, see https://github.com/nervosnetwork/ckb-vm/actions/runs/24328961143/job/71030179557?pr=512

mohanson: Or should we do it the old way: not specifying a compiler on Unix, but specifying a compiler on Windows?

eval-exec: How about create `.cargo/Cargo.toml`, config `CC=clang` in `[env]`?
https://doc.rust-lang.org/cargo/reference/config.html#env

> Or should we do it the old way: not specifying a compiler on Unix, but specifying a compiler on Windows?

Both are fine.

mohanson: Not a good idea. On Windows, `clang` is simply called `clang`, but on Unix, `clang` is usually named `clang-19`, `clang-20`, `clang-21`, and so on. This is why I initially introduced the `clang-finder` package.

eval-exec: `cc-rs` support: `<var>_<target>` - for example:
- CC_x86_64_pc_windows_msvc
- CC_x86_64_unknown_linux_gnu

https://docs.rs/cc/latest/cc/

eval-exec: Does ckb-vm have min/max version requirements for `clang`?

mohanson: There are no restrictions on clang versions.

eval-exec (approved):

XuJiandong (approved):

PR #519, opened 2026-05-29
SC.W / SC.D violate RISC-V A-extension semantics by always performing the store

mohanson:


mohanson:
Since the A extension is not enabled, fixing it seems pointless.

PR #529, opened 2026-07-06
Remove slow path

mohanson:
The slowpath technique was introduced by designing and implementing V Extension, and it's now dead code.


XuJiandong:
remove following：
// Right now we are only dealing with fastpath opcodes, later we might
// (or might not?) expand this with some opcodes in the slowpath category.

XuJiandong (dismissed):

XuJiandong (approved):

PR #530, opened 2026-07-07
Remove A extension

mohanson:
Merging is required after https://github.com/nervosnetwork/ckb-vm/pull/529


XuJiandong:
- Makefile:40 still runs cargo +nightly fuzz run isa_a, but fuzz/fuzz_targets/isa_a.rs and its [[bin]] entry in fuzz/Cargo.toml were deleted. This breaks make fuzz (the isa_a target no longer exists). Line 40 should be removed.

XuJiandong:
src/instructions/execute.rs:1345 has a stale comment mentioning "opcodes in the slowpath category" — the slowpath no longer exists.

XuJiandong (approved):

PR #533, opened 2026-08-12
misc: open unnecessary_cast check in clippy

mohanson:


src/instructions/register.rs
```diff
@@ -334,11 +334,11 @@ impl Register for u32 {
     }
 
     fn rol(&self, rhs: &u32) -> u32 {
-        (*self as u32).rotate_left(*rhs) as u32
+        (*self).rotate_left(*rhs)
```
XuJiandong, line 337: *self.rotate_left(...) would suffice

src/instructions/register.rs
```diff
@@ -334,11 +334,11 @@ impl Register for u32 {
...
     }
 
     fn ror(&self, rhs: &u32) -> u32 {
-        (*self as u32).rotate_right(*rhs) as u32
+        (*self).rotate_right(*rhs)
```
XuJiandong, line 341: *self.rotate_left(...) would suffice

XuJiandong (approved):

src/instructions/register.rs
mohanson: image

PR #536, opened 2026-08-20
Add AGENTS.md

mohanson:


AGENTS.md
```diff
@@ -0,0 +1,44 @@
...
+# Clippy check
+make clippy
+
+# Full CI-level checks
+make ci
+```
+
+## Working Principles
+
+0. Keep changes focused, prioritize fixing the root cause, and avoid unrelated refactors, formatting changes, or dependency upgrades.
```
XuJiandong: `Dependency upgrades` should be explicitly requested. They cannot be done silently.

XuJiandong:
We should write some rules about consensus. For example, cycles and syscall parameters are part of consensus.

Also, add a hardfork rule: if the user specifies the hardfork flag, the agent can break consensus.

XuJiandong (approved):

